PR #2, opened 2023-09-05
Create variables.yml

Chris-green-stfc:
azimuth config file


anish-mudaraddi:
closing this as we've changed how we use argocd

PR #3, opened 2023-09-08
Created appofapps application.yaml

RJaliliSTFC:
And guestbook&rabbit-exporter yaml files
Created folder structure for using appofapps

### DIFF
--- a/rabbit-exporter/application.yaml
+++ b/rabbit-exporter/application.yaml
@@ -5,19 +5,16 @@ metadata:
   namespace: argocd
 spec:
   project: default
-  source:
+  sources:
+  - repoURL: 'https://github.com/prometheus-community/helm-charts.git'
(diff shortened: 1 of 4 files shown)


rabbit-exporter/application.yaml
```diff
@@ -0,0 +1,29 @@
...
+    helm:
+      releaseName: rabbitmq-exporter
+      values: |
+        service:
+          type: LoadBalancer
+
+        rabbitmq:
+          url: https://hv746.nubes.rl.ac.uk:15672/
+          user: monitoring
+          existingPasswordSecret: rabbitmq-exporter-secret
```
DavidFair, line 20: Can we get this to use the values files instead?
Since we're going to want a prod values and dev values

We should also add target revision back and get it working

DavidFair (approved):

PR #5, opened 2023-10-02
Creating the helm structure for argocd

RJaliliSTFC:
Added template yaml files for each application


Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
DavidFair, line 24: We should drop this, since we're not using it
But I'm not sure if helm requires it?

templates/thanos.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: Application
+metadata:
+  name: thanos-{{ $.Values.global.env }}
+  namespace: argocd
+spec:
+  project: {{ $.Values.global.spec.project }}
+  sources:
+  - repoURL: 'https://github.com/bitnami/charts.git'
+    path: bitnami/thanos
+    targetRevision: HEAD
```
DavidFair, line 11: We should have a specific SHA / tag

values/harbor-values.yaml
```diff
@@ -0,0 +1,6 @@
+expose:
+    type: loadBalancer
+    tls:
+        enabled: false
+loadBalancer:
+    IP: "130.246.214.222"
```
DavidFair, line 6: We should drop this IP for the moment, as it's not specific to anything

DavidFair (changes requested):

Chart.yaml
RJaliliSTFC: This is needed

templates/thanos.yaml
RJaliliSTFC: resolved in 4742c4ee91df7b911bfcf411c6d336050efbc76d

values/harbor-values.yaml
RJaliliSTFC: resolved in 4742c4ee91df7b911bfcf411c6d336050efbc76d

DavidFair (approved):

PR #10, opened 2023-11-10
Update harbor-values.yaml to use S3

Koldstream:
Related issue: https://stfc.atlassian.net/jira/software/c/projects/CLDSCRM/boards/77?selectedIssue=CLDSCRM-591

Changes:
Changed config to use S3 instead of Swift, as future versions of Harbor (3.0+) will start to move away from supporting Swift (see: https://github.com/goharbor/harbor/issues/19317)

Commented out configuration left in for when it comes to performance testing.

### DIFF
--- a/prod-values/harbor-values.yaml
+++ b/prod-values/harbor-values.yaml
@@ -41,6 +41,8 @@ jobservice:
 
 # The persistence is enabled by default and a default StorageClass
 # is needed in the k8s cluster to provision volumes dynamically.
+# The persistence is enabled by default and a default StorageClass
+# is needed in the k8s cluster to provision volumes dynamically.
 # Specify another StorageClass in the "storageClass" or set "existingClaim"
 # if you already have existing persistent volumes to use
 #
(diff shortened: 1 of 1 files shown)


prod-values/harbor-values.yaml
```diff
@@ -39,6 +39,8 @@ jobservice:
     jobLoggers: 
         - database
 
+# The persistence is enabled by default and a default StorageClass
+# is needed in the k8s cluster to provision volumes dynamically.
```
meoflynn, line 43: I think this might duplicate the two lines below?

meoflynn (changes requested):

prod-values/harbor-values.yaml
Koldstream: Why yes they are indeedy.

prod-values/harbor-values.yaml
```diff
@@ -50,39 +50,40 @@ persistence:
...
+            secure: true
+            skipverify: false
+            #v4auth: true
+            #chunksize: "5242880"
+            #rootdirectory: /s3/object/name/prefix
+            #storageclass: STANDARD
+            #multipartcopychunksize: "33554432"
+            #multipartcopymaxconcurrency: 100
+            #multipartcopythresholdsize: "33554432"
+            #loglevel: debug
```
DavidFair, line 89: Can we remove these fields if they're not used for our s3 gateways?

prod-values/harbor-values.yaml
```diff
@@ -50,39 +50,40 @@ persistence:
...
+            # Set an existing secret for S3 accesskey and secretkey
+            # keys in the secret should be REGISTRY_STORAGE_S3_ACCESSKEY and REGISTRY_STORAGE_S3_SECRETKEY for registry
+            # existingSecret: ""
+            #region: us-west-1
+            bucket: bucket-name
+            accesskey: 
+            secretkey:
+            regionendpoint: ceph-gw5.gridpp.rl.ac.uk
+            encrypt: true
+            #keyid: mykeyid
```
DavidFair, line 79: Is this required or optional? If it's related to the secret could we put a comment as such

prod-values/harbor-values.yaml
```diff
@@ -50,39 +50,40 @@ persistence:
...
-      #tempurlcontainerkey: false
-      #tempurlmethods:
+        # Specify the type of storage: "filesystem", "azure", "gcs", "s3", "swift",
+        # "oss" and fill the information needed in the corresponding section. The type
+        # must be "filesystem" if you want to use persistent volumes for registry
+        type: s3
+        s3:
+            # Set an existing secret for S3 accesskey and secretkey
+            # keys in the secret should be REGISTRY_STORAGE_S3_ACCESSKEY and REGISTRY_STORAGE_S3_SECRETKEY for registry
+            # existingSecret: ""
```
DavidFair, line 72: Feel free to hardcode a secret name here, and we can use that in the setup process

E.g. "create a secret called `s3-harbor-creds` containing..." in the docs

DavidFair (changes requested):

prod-values/harbor-values.yaml
Koldstream: I'll have a look on how to create a key for both of them.

prod-values/harbor-values.yaml
Koldstream: The uncommented ones are required, the accesskey and secretkey are generated when you make your ec2 creds.

prod-values/harbor-values.yaml
Koldstream: Sure can remove them. I left them just in case we wanted to play with them in performance testing.

prod-values/harbor-values.yaml
```diff
@@ -50,39 +50,27 @@ persistence:
...
+        # backends which not supported it (such as using minio for `s3` storage type), please disable
+        # it. To disable redirects, simply set `disableredirect` to `true` instead.
+        # Refer to
+        # https://github.com/distribution/distribution/blob/main/docs/configuration.md#redirect
+        # for the detail.
+        disableredirect: false
+        # Specify the "caBundleSecretName" if the storage service uses a self-signed certificate.
+        # The secret must contain keys named "ca.crt" which will be injected into the trust store
+        # of registry's containers.
+        # caBundleSecretName:
```
DavidFair, line 63: Nit: I think we can drop this comment completely since we're not using that field

meoflynn (approved):

PR #14, opened 2024-04-03
Staging Cluster Support

DavidFair:
Collab between me and @anish-mudaraddi


clusters/staging-management-cluster/infra-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - name: staging-worker
+    disableAutomated: false
+    chartName: capi
+    namespace: clusters
+    additionalValueFiles:
+      - clusters/staging-worker/overrides/infra/deployment.yaml
+
+    # (optional) can define the git SHA to sync against specific to the app
+    # This overrides global.spec.targetRevision
+    # targetRevision: main
```
DavidFair, line 26: @anish-mudaraddi Can we drop this, and just have one globally pointing to main and per cluster overrides by convention? Since we have a few of these spread around

RJaliliSTFC (approved):

PR #24, opened 2024-04-15
ENH: add longhorn managed application

anish-mudaraddi:
add longhorn as a managed app for argocd 

add config to setup cluster `akhil-harbor-test` as a test cluster for longhorn

### DIFF
--- a/charts/apps/longhorn/values.yaml
+++ b/charts/apps/longhorn/values.yaml
@@ -9,7 +9,8 @@ longhorn:
     enabled: "true"
     ingressClassName: nginx
     host: longhorn.example.com
-    tls: false
+    tls: true
+    tlsSecret: tls-keypair
     secureBackends: false
     path: /
 
--- a/charts/apps/longhorn/values.yaml
+++ b/charts/apps/longhorn/values.yaml
@@ -24,6 +24,7 @@ longhorn:
       longhorn.demo.io/longhorn-storage-node: "true"
 
   defaultSettings:
+    taintToleration: "nvidia.com/gpu:NoSchedule"
     snapshotMaxCount: 10
     snapshotDataIntegrity: true
     snapshotDataIntegtrityCronjob: true
@@ -32,5 +33,7 @@ longhorn:
     allowVolumeCreationWithDegradedAvailability: true
 
   persistence:
+    defaultClassReplicaCount: 3
+    defaultDataLocality: disabled
     migratable: "true"
     
--- a/charts/apps/longhorn/values.yaml
+++ b/charts/apps/longhorn/values.yaml
@@ -3,10 +3,6 @@ longhorn:
   preUpgradeChecker:
     jobEnabled: false
 
-  service:
-    ui:
-      type: ClusterIP
-
   ingress:
     enabled: "true"
     ingressClassName: nginx
--- a/charts/apps/longhorn/values.yaml
+++ b/charts/apps/longhorn/values.yaml
@@ -8,7 +8,6 @@ longhorn:
     ingressClassName: nginx
     host: longhorn.example.com
     tls: true
-    tlsSecret: tls-keypair
     secureBackends: false
     path: /
 
--- a/charts/apps/longhorn/values.yaml
+++ b/charts/apps/longhorn/values.yaml
@@ -1,0 +1,34 @@
+longhorn:
+  preUpgradeChecker:
+    jobEnabled: false
+  service:
+    ui:
+      type: ClusterIP
+
+  ingress:
+    enabled: "true"
+    ingressClassName: nginx
+    host: longhorn.example.com
+    tls: false
+    secureBackends: false
+    path: /
+
+  serviceMonitor:
+    enabled: true
+
+
+  longhornManager:
+    nodeSelector: 
+      longhorn.demo.io/longhorn-storage-node: "true"
+
+  defaultSettings:
+    snapshotMaxCount: 10
+    snapshotDataIntegrity: true
+    snapshotDataIntegtrityCronjob: true
+    replicaAutoBalance: true
+    autoDeletePodWhenVolumeDetachedUnexpectedly: true
+    allowVolumeCreationWithDegradedAvailability: true
+
+  persistence:
+    migratable: "true"
+    
--- a/charts/apps/longhorn/values.yaml
+++ b/charts/apps/longhorn/values.yaml
@@ -1,6 +1,8 @@
 longhorn:
+
   preUpgradeChecker:
     jobEnabled: false
+
   service:
     ui:
       type: ClusterIP
@@ -16,7 +18,6 @@ longhorn:
 
   serviceMonitor:
     enabled: true
-
 
   longhornManager:
     nodeSelector: 
--- a/charts/argocd-apps/values.yaml
+++ b/charts/argocd-apps/values.yaml
@@ -18,3 +18,5 @@ global:
 # NOTE: starts at the root level of the repo
 
 valueFiles:
+  longhorn:
+  - charts/apps/longhorn/values.yaml
--- a/clusters/akhil-harbor-test/argocd-values.yaml
+++ b/clusters/akhil-harbor-test/argocd-values.yaml
@@ -1,0 +1,3 @@
+argo-cd:
+  global:
+    domain: argocd.akhil-harbor-test.com
--- a/clusters/akhil-harbor-test/overrides/infra/deployment.yaml
+++ b/clusters/akhil-harbor-test/overrides/infra/deployment.yaml
@@ -8,6 +8,8 @@ openstack-cluster:
 
   nodeGroupDefaults:
     machineFlavor: l3.tiny
+    nodeLabels:
+      longhorn.demo.io/longhorn-storage-node: true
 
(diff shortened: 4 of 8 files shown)


charts/apps/longhorn/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+  service:
+    ui:
+      type: ClusterIP
+
+  ingress:
+    enabled: "true"
+    ingressClassName: nginx
+    host: longhorn.example.com
+    tls: true
+    tlsSecret: tls-keypair
```
DavidFair, line 15: We haven't come with a convention yet - but our existing services have ingress on with TLS, but don't name a secret
Is it possible to get longhorn to self-sign so it makes initial deployments easy?

charts/apps/longhorn/values.yaml
```diff
@@ -0,0 +1,36 @@
+longhorn:
+
+  preUpgradeChecker:
+    jobEnabled: false
+
+  service:
+    ui:
+      type: ClusterIP
```
DavidFair, line 8: Does this not conflict with ingress enabled? I would have thought it needs to be `service.ui.type = Ingress`

charts/apps/longhorn/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+    nodeSelector: 
+      longhorn.demo.io/longhorn-storage-node: "true"
+
+  defaultSettings:
+    snapshotMaxCount: 10
+    snapshotDataIntegrity: true
+    snapshotDataIntegtrityCronjob: true
+    replicaAutoBalance: true
+    autoDeletePodWhenVolumeDetachedUnexpectedly: true
+    allowVolumeCreationWithDegradedAvailability: true
```
DavidFair, line 32: We need to add some extra fields to our defaults - the `taintToleration` and `persistence` defaults:

https://github.com/stfc/ansible-jupyter/blob/main/roles/deploy_hub/tasks/main.yml#L35

charts/argocd-apps/values.yaml
```diff
@@ -18,3 +18,5 @@ global:
 # NOTE: starts at the root level of the repo
 
 valueFiles:
+  longhorn:
+  - charts/apps/longhorn/values.yaml
```
DavidFair, line 22: Could we check if helm + Argo implicitly add this?
I'd like to do it like additional values files, since IIRC the convention is values.yaml is implied

clusters/akhil-harbor-test/argocd-values.yaml
```diff
@@ -0,0 +1,3 @@
+argo-cd:
+  global:
+    domain: argocd.akhil-harbor-test.com
```
DavidFair, line 3: Needs to be `argocd.akhil-harbor-test.example.com` - this could be a real domain we're squatting

clusters/akhil-harbor-test/overrides/infra/deployment.yaml
```diff
@@ -8,6 +8,8 @@ openstack-cluster:
 
   nodeGroupDefaults:
     machineFlavor: l3.tiny
+    nodeLabels:
+      longhorn.demo.io/longhorn-storage-node: true
```
DavidFair, line 12: @anish-mudaraddi Is there a way to add this to all nodes automatically? Since all members of a cluster are considered longhorn candidates by convention, rather than us having "storage nodes"

docs/DEPLOYING_APPS.md
```diff
@@ -68,5 +68,56 @@ argo-cd:
...
+
+Make sure you have labelled your nodes so that longhorn can use them as storage nodes. You want to label your worker nodes, the default label is `longhorn.demo.io/longhorn-storage-node=true` but you can change this in the cluster-specific values like so:
+
+```
+longhorn:
+  longhornManager:
+    nodeSelector: 
+      # change this to whatever label you want
+      longhorn.demo.io/longhorn-storage-node: "true"
+```
```
DavidFair, line 91: Ditto about using all nodes, I'd make this optional if people want storage nodes
But since we run NVME everywhere there's no advantage having dedicated longhorn targets

docs/DEPLOYING_APPS.md
```diff
@@ -68,5 +68,56 @@ argo-cd:
...
+
+```
+openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout privateKey.key -out certificate.crt
+```
+
+and create the secret like so:
+
+```
+kubectl create secret tls tls-keypair --key /path/to/privateKey.key --cert /path/to/certificate.crt -n longhorn-system
+```
```
DavidFair, line 119: If you deploy longhorn with no `tlsSecret` does it auto-gen the key for you? Most charts do hence asking

docs/DEPLOYING_APPS.md
```diff
@@ -68,5 +68,56 @@ argo-cd:
...
+
+# Post-deployement steps
+
+None - it should just work
```
DavidFair, line 123: Optionally they might want to change their default storage class to longhorn from Cinder

DavidFair (changes requested):
Nice work, some nit picks but I'm keen to test this

clusters/akhil-harbor-test/overrides/infra/deployment.yaml
DavidFair: (Sorry for the at, I thought this was Akhil's PR)

charts/apps/longhorn/values.yaml
anish-mudaraddi: don't think that's possible from a look in the docs or the helm values

anish-mudaraddi: You can easily turn off ingress if you want easy deployments

charts/apps/longhorn/values.yaml
anish-mudaraddi: No? 
The service is a clusterIP and then we define an ingress resource to tie this ip to ingress. 
I can remove this since its a default though

clusters/akhil-harbor-test/argocd-values.yaml
anish-mudaraddi: ignore this since we'll delete this in a follow-up PR

clusters/akhil-harbor-test/overrides/infra/deployment.yaml
anish-mudaraddi: I'd rather not do that - I don't want to add unnecessary labels by default on nodes

docs/DEPLOYING_APPS.md
anish-mudaraddi: I saw problems with enabling it on control-plane machine - if you want to experiment with enabling on control plane machines do it in a follow up PR

docs/DEPLOYING_APPS.md
anish-mudaraddi: no

docs/DEPLOYING_APPS.md
anish-mudaraddi: well obviously - but if they're deploying longhorn they should know what it does right?

anish-mudaraddi: I don't know how they can do that

charts/apps/longhorn/values.yaml
DavidFair: Lets leave it as self-signed, shame it doesn't allow us to do it

charts/apps/longhorn/values.yaml
DavidFair: Yeah, I want to strip anything we leave as default so it's obvious what we have an opinion on?
That's the way our Kolla deployments on the ops side is going too

docs/DEPLOYING_APPS.md
DavidFair: Oh it shouldn't be on control plane machines, but I thought there was a taint on the control plane which longhorn didn't tolerate?

docs/DEPLOYING_APPS.md
```diff
@@ -120,4 +120,8 @@ kubectl create secret tls tls-keypair --key /path/to/privateKey.key --cert /path
 
 # Post-deployement steps
 
-None - it should just work
\ No newline at end of file
+Longhorn is already setup to be the default storageclass - if you're using CAPI you will need to drop csi-cinder storage class as being the default - you can do this by running
+
+```
+kubectl patch storageclass csi-cinder -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
```
DavidFair, line 126: Nit: Can we not do this through Argo - every patch is basically us working around the config management?

charts/argocd-apps/values.yaml
anish-mudaraddi: it doesn't - I'll fix this later

docs/DEPLOYING_APPS.md
anish-mudaraddi: yea we should use argo for this - I'll fix this

RJaliliSTFC (approved):

PR #25, opened 2024-04-16
Adding victoria metrics test cluster

RJaliliSTFC:


clusters/victoria-metrics-test/deployment.yaml
```diff
@@ -0,0 +1,23 @@
...
+  addons:
+    ingress:
+      enabled: true
+      nginx:
+        release:
+          values:
+            controller:
+              service:
+                # Temp pending move to test.harbor.stfc.ac.uk
+                loadBalancerIP: "130.246.211.115"
```
meoflynn, line 23: Is this correct for the cluster ingress as the comment is for test harbor?

meoflynn (changes requested):

anish-mudaraddi (approved):

meoflynn (approved):

PR #26, opened 2024-04-16
ENH: Enable Monitoring on CAPI

anish-mudaraddi:
Enable KubePrometheusStack Addon for CAPI and configure emailing config
Tested on my child cluster `anish-monitoring-test`

You may notice that `charts/infra/capi/addons/monitoring.yaml` and `clusters/anish-monitoring-test/infra/deployment.yaml` have very similar config - this is because we cannot get `cloud-deployed-infra` to pick up the new values file for capi `charts/infra/capi/addons/monitoring.yaml` without pushing that file to `main` branch. 

TLDR: when reviewing this PR ignore all files in `clusters/anish-monitoring-test/` as they will be deleted anyway

 can't add a new file for capi as we are testing on a child cluster and cannot modify cloud-deployed-infra to point to our test branch so we have duplicated our monitoring config as cluster-specific config for our test cluster

### DIFF
--- a/charts/infra/capi/addons/monitoring.yaml
+++ b/charts/infra/capi/addons/monitoring.yaml
@@ -7,6 +7,12 @@ openstack-cluster:
       kubePrometheusStack:
         release:
           values:
+
+            defaultRules:
+              additionalRuleLabels:
+                cluster: anish-monitoring-cluster
+                env: dev
+            
             alertmanager:
               enabled: true
               service:
@@ -34,7 +40,6 @@ openstack-cluster:
                   smtp_require_tls: false
 
                 route:
-                  group_by: ['cluster', 'service']
                   group_wait: 30m
                   group_interval: 30m
                   repeat_interval: 2d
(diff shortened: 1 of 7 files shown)


RJaliliSTFC (approved):

charts/infra/capi/addons/monitoring.yaml
```diff
@@ -0,0 +1,183 @@
...
+            alertmanager:
+              enabled: false
+              service:
+                type: ClusterIP
+              
+              ingress:
+                ingressClassName: nginx
+                enabled: true
+                hosts: 
+                  -  alertmanager.example.com
```
DavidFair, line 25: Ditto about these, since they're per-cluster values

docs/DEPLOYING_INFRA.md
```diff
@@ -109,8 +184,12 @@ spec:
...
       - name: openstack-cluster.apiServer.floatingIP
         value: <APIServer floating ip here>
-      - name: openstack-cluster.addons.ingress.nginx.release.values.controller.service.loadbalancerIP
-        value: <nginx ingress controller floating here>
+      - name: >-
+          openstack-cluster.addons.monitoring.kubePrometheusStack.release.values.alertmanager.config.global.smtp_smarthost
+        value: <hostname>:<port>
+      - name: >-
+          openstack-cluster.addons.monitoring.kubePrometheusStack.release.values.alertmanager.config.receivers[1].email_configs[0].to
+        value: <email-address>
```
DavidFair, line 192: I thought this was in the config, since it's a "public" email anyway?

anish-mudaraddi: if you want to send alerts somewhere other than cloud-support - you set it here. We want that to be a secret so we don't expose other emails

PR #31, opened 2024-04-16
DOC: New documentation

RJaliliSTFC:
doc detailing deployment of new cluster in steps

### Description:
Created a new doc to detail steps for creating a new cluster off of staging

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


docs/DeployNewCluster.md
```diff
@@ -0,0 +1,109 @@
+# Deploying a Worker Cluster from the Staging Cluster
```
DavidFair, line 1: `Nit: from an existing cluster`

Since this will also work for prod too

docs/DeployNewCluster.md
```diff
@@ -0,0 +1,109 @@
...
+openstack-cluster:
+  controlPlane:
+    machineCount: 3
+
+  nodeGroups:
+    - name: default-md-0
+      machineCount: 2
+
+  nodeGroupDefaults:
+    machineFlavor: l3.tiny
```
DavidFair, line 43: Can we remove these? If people don't have opinions we should use the defaults in the chart

It's only at 3 + 2 because I wanted to check if Argo could handle scaling the kcp and workers for example

docs/DeployNewCluster.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+## 8. Once these changes are merged then the staging cluster will start to spin up your new cluster
+
+You will then be able to retrieve the kubeconfig for the new cluster once it is spun up
+
+Use `clusterctl describe cluster $CLUSTER_NAME -n clusters` to check the status of the cluster
+
+Use `clusterctl get kubeconfig $CLUSTER_NAME -n clusters > $CLUSTER_NAME.kubeconfig` to output the kubeconfig for the cluster
+
```
DavidFair, line 109: Copy this to a new VM to continue working. You may want to read <deploying_apps link> if you're planning on creating a new app on your cluster

DavidFair (commented):
Some nits and tidy up

Also I want to sign-post people on something they'd naturally do next, creating an app on the cluster they just created

docs/DeployNewCluster.md
RJaliliSTFC: I thought the word staging meant where we 'stage' clusters not the distinction from prod/dev, ideally it would be the prod-staging and dev-staging clusters

docs/DeployNewCluster.md
RJaliliSTFC: I personally wanted to keep them in there as being someone who wants to think about node/worker count, having it shown where to change already is helpful

docs/DeployNewCluster.md
RJaliliSTFC: I was unsure about whether to do the ./deploy cluster-name step on this doc as then it would be 'creating a new cluster + deploying argo to it' in one doc, although when you look at the steps needed to then deploy argo it's not enough to justify a full doc, was hoping for some opinions on this

RJaliliSTFC:
Planning on going over the Read-me as well, as I think it's also  a bit unstructured. Planning on having it read more like Title: doc for step. A bit more explanation for each link.

docs/DeployNewCluster.md
DavidFair: Ah but you see how it's already overloaded - maybe `Deploying a worker cluster from a management cluster`?

docs/DeployNewCluster.md
DavidFair: It up to you, if it flows better a few duplicated lines are not problem - getting someone to change page for a 1-2 lines is worth the duplication
It's when entire paragraphs get duplicated it's a PITA to keep updated

RJaliliSTFC: Going to write some duplicate steps that flow into adding apps doc

DavidFair: Also for the worker case might be worth giving them this one-liner to helpfully copy from one VM to another (assuming they have Linux locally)

- Copy the kubeconfig directly from host A to B (using via the current SSH session), if you don't have agent forwarding
scp -3 user@mgmt-ip:/tmp/worker.kubeconfig user@new-vm:/tmp


As I find myself using it a lot and I imagine people with WSL will too

docs/DeployArgoandApps.md
```diff
@@ -0,0 +1,118 @@
...
+
+## 2. Content of Chart.yaml
+The app-chart-repository is the location where the helm chart is hosted
+```
+apiVersion: v2
+name: <app-name>
+version: 1.0.0
+dependencies:
+- name: <app-name>
+  repository: <app-chart-repository>
```
anish-mudaraddi, line 58: you should also say to set the version here

docs/DeployArgoandApps.md
```diff
@@ -0,0 +1,118 @@
...
+ 
+   #name of the chart to install - corresponds with chart name in charts/apps/
+  chartName: argocd
+ 
+   #namespace where the chart contents will be installed to
+  namespace: argocd
+ 
+   #(optional) can define the git SHA to sync against specific to the app
+   #This overrides global.spec.targetRevision
+  targetRevision: main
```
anish-mudaraddi, line 106: I'd remove this to avoid confusion - most people won't need to include this unless your messing with capi stuff

docs/DeployNewCluster.md
```diff
@@ -0,0 +1,140 @@
...
+      # change this if you're using a fork
+      repoURL: https://github.com/stfc/cloud-deployed-apps.git 
+
+      # namespace where the argocd applications will be installed to 
+      namespace: argocd
+```
+
+## 6. Contents for argocd-values.yaml
+### The domain name is set by you, if you are testing a good name to use is "test.argocd.stfc.nubes.ac.uk" and you will have to add it to your etc hosts link for this
+`https://www.wikihow.com/Edit-the-Hosts-File-on-Windows#:~:text=1%20Open%20the%20Start%20menu%2C%20and%20click%20%22Computer%22.,the%20file%20directly%20to%20your%20%22etc%22...%20See%20More`
```
anish-mudaraddi, line 83: Edit this line like so:
```
[Editing Hosts file on Windows](https://www.wikihow.com/Edit-the-Hosts-File-on-Windows#:~:text=1%20Open%20the%20Start%20menu%2C%20and%20click%20%22Computer%22.,the%20file%20directly%20to%20your%20%22etc%22...%20See%20More)
```

anish-mudaraddi (changes requested):

docs/DeployArgoandApps.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+## 2. Content of Chart.yaml
+The app-chart-repository is the location where the helm chart is hosted
+```
+apiVersion: v2
+name: <app-name>
+version: 1.0.0
+dependencies:
+- name: <app-name>
```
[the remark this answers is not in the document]
RJaliliSTFC: Resolved

docs/DeployArgoandApps.md
RJaliliSTFC: Resolved

docs/DeployNewCluster.md
RJaliliSTFC: Resolved

anish-mudaraddi (approved):

DavidFair (approved):

PR #39, opened 2024-04-19
Victoria Metrics - Initial Work

DavidFair:
### Description:

Cleaned up history of @RJaliliSTFC 's Victoria metrics work
This:

- Creates a Victoria Metrics App
- Backs it using the Longhorn Storage
- Deploys an initial version to a test cluster

I've:
- Cleaned up / squashed the existing commit history
- Follow-up from review comments and my own

### Special Notes:

- I'll do a tidy-up commit for the things I noticed reviewing, without adding any new work.
- Once it's been reviewed and we're happy with the changes, we'll follow up with a deployment PR to add it to prod / staging clusters
- We need documentation but @RJaliliSTFC is not available at the time of writing. I'll follow up with him where it's currently up to

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/charts/apps/victoria-metrics/values.yaml
+++ b/charts/apps/victoria-metrics/values.yaml
@@ -1,0 +1,94 @@
+# Values for victoria-metrics.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+vmselect:
+  horizontalPodAutoscaler:
+    # -- Use HPA for vmselect component
+    enabled: true
+    # -- Maximum replicas for HPA to use to to scale the vmselect component
+    maxReplicas: 7
+    # -- Minimum replicas for HPA to use to scale the vmselect component
+    minReplicas: 3
+    # -- Metric for HPA to use to scale the vmselect component
+    metrics:
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 50
+
+  service:
+    type: LoadBalancer
+
+  ingress:
+    # -- Enable deployment of ingress for vmselect component
+    enabled: false
+    # -- Ingress annotations
+    annotations: {}
+    #   kubernetes.io/ingress.class: nginx
+    #   kubernetes.io/tls-acme: 'true'
+
+vminsert:
+  # Horizontal Pod Autoscaling
+  horizontalPodAutoscaler:
+    # -- Use HPA for vminsert component
+    enabled: true
+    # -- Maximum replicas for HPA to use to to scale the vminsert component
+    maxReplicas: 10
+    # -- Minimum replicas for HPA to use to scale the vminsert component
+    minReplicas: 2
+    # -- Metric for HPA to use to scale the vminsert component
+    metrics:
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 50
+
+  service:
+    type: LoadBalancer
+
+  ingress:
+    # -- Enable deployment of ingress for vminsert component
+    enabled: false
+    # -- Ingress annotations
+    annotations: {}
+    #   kubernetes.io/ingress.class: nginx
+    #   kubernetes.io/tls-acme: 'true'
+    extraLabels: {}
+    # -- Array of host objects
+    hosts: []
+    # - name: vminsert.local
+    #   path: /insert
+    #   port: http
+    # -- Array of TLS objects
+    tls: []
+    #   - secretName: vminsert-ingress-tls
+    #     hosts:
+    #       - vminsert.local
+    # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
+    # See https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress
+    # ingressClassName: nginx
+    # -- pathType is only for k8s >= 1.1=
+    pathType: Prefix
+
+vmstorage:
+  podDisruptionBudget:
+    enabled: false
+    # minAvailable: 1
+    # maxUnavailable: 1
+    labels: {}
+  # -- Data retention period. Supported values 1w, 1d, number without measurement means month, e.g. 2 = 2month
+  retentionPeriod: 1
+  replicaCount: 3
+
+  persistentVolume:
+    enabled: false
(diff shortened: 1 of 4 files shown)


charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+      # -- Minimum replicas for HPA to use to scale the vmselect component
+      minReplicas: 3
+      # -- Metric for HPA to use to scale the vmselect component
+      metrics:
+        - type: Resource
+          resource:
+            name: cpu
+            target:
+              type: Utilization
+              averageUtilization: 50
```
DavidFair, line 20: I'm going to disable the autoscaler, but leave the definition here for the future
Autoscaling makes sense where we spin down nodes, but since we're going to have dedicated VMs we might as well use the full capacity of them if we have them

charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+    service:
+      type: LoadBalancer
+  
+    ingress:
+      # -- Enable deployment of ingress for vmselect component
+      enabled: false
+      # -- Ingress annotations
+      annotations: {}
+      #   kubernetes.io/ingress.class: nginx
+      #   kubernetes.io/tls-acme: 'true'
```
DavidFair, line 31: I'm going to update this to use ingress, as per the other things
I'll do a different PR to get cert-bot installed too as I need this for some other work

charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+      metrics:
+        - type: Resource
+          resource:
+            name: cpu
+            target:
+              type: Utilization
+              averageUtilization: 50
+  
+    service:
+      type: LoadBalancer
```
DavidFair, line 52: Ditto ^

clusters/victoria-metrics-test/argocd-values.yaml
```diff
@@ -0,0 +1,4 @@
+argo-cd:
+  global: 
+    # default is argocd.example.com
+    domain: "argocd.victoria-metrics.example.com"
```
anish-mudaraddi, line 4: change this to argocd.victoria-metrics.nubes.stfc.ac.uk or something similar - I don't like example.com :)

charts/apps/victoria-metrics/values.yaml
anish-mudaraddi, line 52: This should be using ingress not loadbalancer surely?

anish-mudaraddi: Maybe @RJaliliSTFC hasn't done that yet?

anish-mudaraddi (changes requested):

charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+      pathType: Prefix
+  
+  vmstorage:
+    podDisruptionBudget:
+      enabled: false
+      # minAvailable: 1
+      # maxUnavailable: 1
+      labels: {}
+    # -- Data retention period. Supported values 1w, 1d, number without measurement means month, e.g. 2 = 2month
+    retentionPeriod: 1
```
DavidFair, line 85: Even though the comment explains it I'm going to formally put this as `1m`

charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+    replicaCount: 3
+  
+    persistentVolume:
+      enabled: true
+      # --  Existing Claim name. Requires vmstorage.persistentVolume.enabled: true. If defined, PVC must be created manually before volume will be bound
+      existingClaim: ""
+  
+      size: 10Gi
```
DavidFair, line 93: I'm going to bump this to say 50GB, since 10GB is pretty small considering the storage on our flavours

charts/apps/victoria-metrics/values.yaml
anish-mudaraddi: he mentioned something like he wanted `victoria-select.prod.nubes.stfc.ac.uk` and `victoria-insert.prod.nubes.stfc.ac.uk` for the domains

clusters/victoria-metrics-test/argocd-values.yaml
DavidFair: Any objections to having this merged as-is
Then doing a deployment follow-up PR to allocate all the DNS names and remove this test instance?

Mostly because we don't own the DNS currently that you're proposing - and I don't want to get into the habit of squatting names we haven't asked for

DavidFair:
There's still a few TODOs we can follow up on w.r.g.t Longhorn + VM storage in a future PR

- Longhorn should have of anti-affinity on the volumes, so they don't land on the same node
- We should disable replication in Longhorn (no point replicating a replica)
- We need to set the dataLocality to best effort, so that we don't have east-west traffic of storage and replicas

Additionally we should look at how many VM nodes we'll deploy and base our storage replicas based on that. But that requires us to do an initial deployment and see how it scales

I'll @RJaliliSTFC since this will also be of interest to him

charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - name: # select.vm.example.com
+          path: /
+          port: http
+      tls:
+        - secretName: vm-ingress-tls
+          hosts:
+            # - select.vm.example.com
+
+  vminsert:
+    replicaCount: 3
```
RJaliliSTFC, line 18: Scale this component up to 5 replicas as well

charts/apps/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+          path: /
+          port: http
+      tls:
+        - secretName: vm-ingress-tls
+          hosts:
+            # - insert.vm.example.com
+
+  vmstorage:
+    retentionPeriod: "1" # month
+    replicaCount: 3
```
RJaliliSTFC, line 34: Still need to find out data for this component, but realistically the more the better for performance/stability. I would set this to 5 by default for the test

anish-mudaraddi (approved):

anish-mudaraddi (approved):

meoflynn (approved):

PR #40, opened 2024-04-19
MAINT: change the name of staging management cluster

anish-mudaraddi:
### Description:

to follow naming convention so its the same as prod-mgmt-cluster

This is a breaking change! - don't merge this until we're ready to run ./deploy again on the staging cluster. This is part of a test we need to document to re-make an existing cluster

Examples of special notes that must be included in the PR:
- If this is a hotfix for production, which needs to be deployed after merging
- If this requires manual work to deploy the PR, e.g. a parameter change
- If this has associated internal documentation too
-->

---

### Submitter:

Have you:

---

### Reviewer

Have you:


DavidFair:
@anish-mudaraddi Do you still want to land this?
I'm happy to re-deploy the cluster, since we had to change the external network UUID recently anyway

PR #41, opened 2024-04-19
ENH: add prod-mgmt cluster

anish-mudaraddi:
### Description:
Add prod management cluster and make it self-managing

### Submitter:

Have you:

---

### Reviewer

Have you:

* [x] Verified this PR uses the correct label(s) based on the rules above?
* [x] Checked if this could affect production (e.g. a global value that's changed without an override)?

### DIFF
--- a/charts/infra/capi/values.yaml
+++ b/charts/infra/capi/values.yaml
@@ -27,7 +27,7 @@ openstack-cluster:
 
   nodeGroupDefaults:
     autoscale: false
-    machineFlavor: l3.micro
+    machineFlavor: l3.nano
 
     healthCheck:
       enabled: true
@@ -45,11 +45,10 @@ openstack-cluster:
   additionalPackages: []
 
   controlPlane:
+    machineCount: 3
     remediationStrategy:
       retryPeriod: 20m0s
       minHealthyPeriod: 1h0m0s
-
-    machineCount: 5
     machineFlavor: l3.nano
 
   kubernetesVersion: "1.28.7"
--- a/charts/infra/capi/values.yaml
+++ b/charts/infra/capi/values.yaml
@@ -27,7 +27,7 @@ openstack-cluster:
 
   nodeGroupDefaults:
     autoscale: false
-    machineFlavor: l3.nano
+    machineFlavor: l3.micro
 
     healthCheck:
       enabled: true
@@ -45,7 +45,7 @@ openstack-cluster:
   additionalPackages: []
 
   controlPlane:
-    machineCount: 3
+    machineCount: 5
     remediationStrategy:
       retryPeriod: 20m0s
       minHealthyPeriod: 1h0m0s
--- a/clusters/prod-mgmt/app-values.yaml
+++ b/clusters/prod-mgmt/app-values.yaml
@@ -2,6 +2,6 @@ global:
   clusterName: prod-mgmt
   spec:
     source:
-      targetRevision: prod
+      targetRevision: main
       repoURL: https://github.com/stfc/cloud-deployed-apps.git
       namespace: argocd
--- a/clusters/prod-mgmt/app-values.yaml
+++ b/clusters/prod-mgmt/app-values.yaml
@@ -2,6 +2,11 @@ global:
   clusterName: prod-mgmt
   spec:
     source:
-      targetRevision: main
+      targetRevision: 000cf30993bd7fb191281ad3e5da7c4ddd95aba6
       repoURL: https://github.com/stfc/cloud-deployed-apps.git
       namespace: argocd
+
+# repoint cloud-deployed-apps application to point to main so we pick up latest changes
+# in case we move global targetRevision 
+cloudDeployedApps:
+  targetRevision: main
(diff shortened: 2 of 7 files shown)


clusters/prod-mgmt/app-values.yaml
```diff
@@ -0,0 +1,7 @@
+global:
+  clusterName: prod-mgmt
+  spec:
+    source:
+      targetRevision: main
```
DavidFair, line 5: This should be a commit SHA for production - since we want it to rollout explicitly rather than be on bleeding edge

clusters/prod-mgmt/infra-values.yaml
```diff
@@ -0,0 +1,15 @@
+global:
+  clusterName: prod-mgmt
+  spec:
+    source:
+      targetRevision: main
```
DavidFair, line 5: Ditto ^ Since we don't want to say have someone merge a new image then all hell break loose on prod before staging has flagged it

clusters/prod-mgmt/overrides/infra/deployment.yaml
```diff
@@ -0,0 +1,57 @@
+openstack-cluster:
+  controlPlane:
+    machineCount: 3
+
+  nodeGroups:
+    - name: default-md-0
+      machineCount: 2
+
+  nodeGroupDefaults:
+    machineFlavor: l3.nano
```
DavidFair, line 10: We should use our upstream defaults as they're the safest config for a production deployment instead of overriding

clusters/prod-mgmt/overrides/infra/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  cloudCredentialsSecretName: prod-mgmt-cloud-credentials
```
DavidFair, line 12: Just to double check - were the app creds created whilst logging in with the dedicated service account on OpenStack?
If not, we need to rotate the app creds to use it. This doesn't require recreating the cluster, since K8s is smart enough to rotate them through the machine templates

clusters/prod-mgmt/app-values.yaml
anish-mudaraddi: set the branch to latest commit now then?

clusters/prod-mgmt/overrides/infra/deployment.yaml
anish-mudaraddi: I've moved it to defaults

clusters/prod-mgmt/overrides/infra/deployment.yaml
anish-mudaraddi: we need to rotate the app-creds with a service account one yea - I've created it with my own app-creds - if you create the secret the same way its mentioned in the docs - argocd will rotate it for us since it manages the secret

clusters/prod-mgmt/app-values.yaml
anish-mudaraddi: I've created a branch called "prod" that it follows

charts/infra/capi/values.yaml
```diff
@@ -27,7 +27,7 @@ openstack-cluster:
 
   nodeGroupDefaults:
     autoscale: false
-    machineFlavor: l3.micro
+    machineFlavor: l3.nano
```
DavidFair, line 30: This need to stay as micro - we've seen it where after deploying the add-ons / CNI / security ...etc. there's no space left on the worker

charts/infra/capi/values.yaml
```diff
@@ -45,11 +45,10 @@ openstack-cluster:
   additionalPackages: []
 
   controlPlane:
+    machineCount: 3
     remediationStrategy:
       retryPeriod: 20m0s
       minHealthyPeriod: 1h0m0s
-
-    machineCount: 5
```
DavidFair, line 52: Why has this been changed from 5->3 
We use 5 in our upstream values: https://github.com/stfc/cloud-capi-values/blob/master/user-values.yaml#L16 for HA by default?

clusters/prod-mgmt/app-values.yaml
```diff
@@ -0,0 +1,7 @@
+global:
+  clusterName: prod-mgmt
+  spec:
+    source:
+      targetRevision: prod
```
DavidFair, line 5: This should be a sha that belongs to main, that way we can rollback to a known good point if something goes wrong based on the SHA rather than a pointer

charts/infra/capi/values.yaml
anish-mudaraddi: you said to make the overrides the defaults in your last commit - I'll change it back then

clusters/prod-mgmt/app-values.yaml
anish-mudaraddi: I'll leave it as main then - we'll make another commit after

DavidFair:
Marking as an enhancement too since we needed some chart changes which should be tested - I'll quickly use a self-managed cluster pointed to this to validate it works

DavidFair (approved):

meoflynn (approved):

PR #43, opened 2024-04-19
Add cert manager

DavidFair:
### Description:
Adds cert manager as a global item, as per the documentation https://cert-manager.io/docs/installation/helm/#installing-cert-manager-as-subchart
Since we can only have one `cloud-deployed-apps` (and only an app would use it) we use it as a dependency.

By default we enable it with the self-signed issuer. This should encourage developers to test with cert-manager and make it easy to switch to a trusted provider later, unifying the dev and prod paths in most cases.

### Special Notes:

- This should be tested against an existing pre-deployed cluster too, by pointing it to this branch. As we should verify that Argo is smart enough to fetch the dependencies, since helm wasn't during `deploy.sh`

---

### Submitter:

Have you:

---

### Reviewer

Have you:


scripts/deploy.sh
```diff
@@ -33,11 +33,15 @@ done
...
   echo "Installing ArgoCD CLI..."
   curl -sSL -o argocd-linux-amd64 https://github.com/argoproj/argo-cd/releases/latest/download/argocd-linux-amd64
-  sudo install -m 555 argocd-linux-amd64 /usr/local/bin/argocd
+  sudo install -m 755 argocd-linux-amd64 /usr/local/bin/argocd
   rm argocd-linux-amd64
 fi
 
 echo "Creating App of Apps for cluster $CLUSTER_NAME..."
+# Required to install cert-manager as a dependency
+helm dependency update ../charts/argocd-apps
```
anish-mudaraddi, line 42: just curious - do you need to do this? Thought argo does this for you

scripts/deploy.sh
```diff
@@ -33,11 +33,15 @@ done
...
+kubectl create namespace cert-manager || true
```
anish-mudaraddi, line 43: argocd should create the namespace for you

charts/argocd-apps/values.yaml
```diff
@@ -3,20 +3,30 @@ global:
...
+
+  le-staging:
+    enabled: false
+
+  le-prod:
+    enabled: false
 
-valueFiles:
-  longhorn:
-  - charts/apps/longhorn/values.yaml
```
anish-mudaraddi, line 22: why'd you get rid of this? 
atm we need to define default values files for all our apps this way - so add this back

anish-mudaraddi (changes requested):
looks fine - few comments

anish-mudaraddi:
maybe think about deploying cert-manager with a standalone application rather than it being tied with cloud-deployed-apps - therefore it will be more stable on prod - look at the way argocd application is defined here - https://github.com/stfc/cloud-deployed-apps/blob/main/charts/argocd-apps/templates/argocd-app.yaml

scripts/deploy.sh
DavidFair: Unfortunately, to get the app installed into Argo CD helm needs the dependencies including cert-manager up-front
Once it's created then yeah Argo does the rest

charts/argocd-apps/values.yaml
DavidFair: It works without this though?
Since our other apps including the one in `setup_harbor_infra` doesn't include these lines but still grabs `values.yaml` values

DavidFair:
> maybe think about deploying cert-manager with a standalone application rather than it being tied with cloud-deployed-apps - therefore it will be more stable on prod - look at the way argocd application is defined here - https://github.com/stfc/cloud-deployed-apps/blob/main/charts/argocd-apps/templates/argocd-app.yaml

https://cert-manager.io/docs/installation/helm/#installing-with-helm

As per the docs we need to...

> Be sure never to embed cert-manager as a sub-chart of other Helm charts; cert-manager manages non-namespaced resources in your cluster and care must be taken to ensure that it is installed exactly once.

As this application can only be installed once, it guards against something akin to:

```
apps:
  - name: foo  # brings the cert-manager dep
  - name: cert-manager
```

We could just move it to an app and guard against it in PRs, but I want to make it so even self-signed certs uses the same mechanism - thoughts?

charts/argocd-apps/values.yaml
anish-mudaraddi: huh, are you 100% sure that works? Guess if a `values.yaml` is next to the `Chart.yaml` it automatically picks it up

anish-mudaraddi:
@DavidFair - see this branch with a commit on how I think you should add cert-manager as a single-use app 
https://github.com/stfc/cloud-deployed-apps/tree/updated-cert-manager

benefits of this: you decouple `cert-manager` from `cloud-deployed-apps` - single responsibility principle. 
It's more stable

DavidFair:
If we do it as a standalone argo app - is there a way to turn an application off or on through it?
I like the idea that we use `cert-manager.enabled` to toggle it at an Argo level rather than the dependency

anish-mudaraddi (approved):

clusters/david-pr-test/app-values.yaml
```diff
@@ -2,6 +2,6 @@ global:
   clusterName: david-pr-test
   spec:
     source:
-      targetRevision: longhorn-update
+      targetRevision: Add_cert_manager
```
anish-mudaraddi, line 5: This will need to be changed back to main - or better yet the test cluster can be cleaned up in a followup PR

charts/argocd-apps/Chart.yaml
RJaliliSTFC, line 1: Are we incrementing the version every time we add a new app?

clusters/david-pr-test/app-values.yaml
DavidFair: I'll do a follow-up PR to remove this

charts/argocd-apps/Chart.yaml
DavidFair: In this case it's because I've changed the chart itself by adding a new file to the `templates` directory, if we change the files in this directory we should do a semver bump

Adding a new app, since it's a different dir, we'd bump the version there instead

PR #45, opened 2024-04-22
BUG: explicitly define externalNetworkID

anish-mudaraddi:
### Description:
if the project has multiple external networks - cluster deployment fails - we need to explicitly define the "id" of the network we want to use for each cluster

**NOTE**: This will require us to re-make all of our clusters because this is a protected field and can't be changed on running clusters 

---

### Submitter:

Have you:

---

### Reviewer

Have you:


meoflynn (approved):

anish-mudaraddi:
related https://github.com/stfc/cloud-capi-values/pull/17

RJaliliSTFC (approved):

PR #48, opened 2024-04-22
Rename secret for victoria-metrics-prod cluster

RJaliliSTFC:
### Description:
change name to be more exact

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


anish-mudaraddi (approved):
approved - but this won't take affect until we move up the prod-tag. 
I'd recommend duplicate secrets - `victoria-metrics-test-cloud-config` and `victoria-metrics-prod-cloud-config` - and once we update the prod-tag it should just use the other

RJaliliSTFC:
Unimportant change for now

PR #52, opened 2024-04-23
make prod clusters follow prod tag

anish-mudaraddi:
commit SHAs don't work - we'll go back to the original - we'll make a prod tag on the latest commit and deploy it off of that - we lose CI - and need to manually run deploy for now


RJaliliSTFC (approved):

DavidFair (approved):
:+1:

We'll give up easy rollbacks for stability and come back to this later on staging

PR #54, opened 2024-04-23
DEPLOY: Add prod-worker cluster

anish-mudaraddi:
just manage infra for now - apps to follow


clusters/prod-worker/argocd-values.yaml
```diff
@@ -0,0 +1,3 @@
+argo-cd:
+  global:
+    domain: argocd-worker.prod.nubes.stfc.ac.uk
```
RJaliliSTFC, line 3: Does this need to be in quotations?

anish-mudaraddi: no need

DavidFair (approved):

PR #55, opened 2024-04-23
Changed cluster name to vm-prod to try handle metadata limit

RJaliliSTFC:
### Description:
Shortened cluster name so that pod metadata labels are <63 characters long

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


anish-mudaraddi (approved):
happy with this  - but it means that you're current cluster will be destroyed - bare that in mind

PR #60, opened 2024-04-24
BUG: fix monitoring intervals

anish-mudaraddi:
monitoring intervals were close together and spamming our ticketing system - alertmanager sending multiple tickets for the same issue should not be a problem anymore

### DIFF
--- a/charts/infra/capi/addons/monitoring.yaml
+++ b/charts/infra/capi/addons/monitoring.yaml
@@ -35,6 +35,7 @@ openstack-cluster:
 
               config:
                 global:
+                  # how long to wait until a resolved message comes in - if an alert has been sent and was subsequently resolved
                   resolve_timeout: 1h
                   smtp_smarthost: 
                   smtp_from:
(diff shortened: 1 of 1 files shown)


charts/infra/capi/addons/monitoring.yaml
```diff
@@ -47,8 +48,18 @@ openstack-cluster:
                       matchers:
                       - alertname = "Watchdog"
                     - receiver: default-receiver
-                      group_wait: 30m
-                      group_interval: 30m
+                      # how long the alert needs to be firing before we send an alert
+                      # this is high so more likely that issues k8s resolves by itself will be ignored
+                      group_wait: 1h
```
DavidFair, line 53: What are we seeing that takes >30 minutes to resolve?
IMO if it's been failing for >30 minutes we should look at it, even if it auto resolves since it potentially impacts the service

anish-mudaraddi: move it back to 30mins?

DavidFair (approved):

meoflynn (approved):

PR #64, opened 2024-05-02
MAINT: setup monitoring on staging-worker

anish-mudaraddi:
### Description:
setup monitoring and use updated floating ip for ingress as the previous floating ip doesn't seem to exist anymore

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/clusters/staging-worker/overrides/infra/deployment.yaml
+++ b/clusters/staging-worker/overrides/infra/deployment.yaml
@@ -19,5 +19,38 @@ openstack-cluster:
           values:
             controller:
               service:
-                # Temporary for testing DR scenario
-                loadBalancerIP: "130.246.211.226"
+                loadBalancerIP: "130.246.211.106"
+    
+    kubePrometheusStack:
+      release:
+        values:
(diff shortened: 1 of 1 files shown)


clusters/staging-worker/overrides/infra/deployment.yaml
```diff
@@ -19,5 +19,38 @@ openstack-cluster:
...
+                  - prometheus-worker.staging.nubes.stfc.ac.uk
+                  secretName: tls-keypair
+          grafana:
+            ingress:
+              hosts:
+                - grafana-worker.staging.nubes.stfc.ac.uk
+              tls:
+                - hosts: 
+                  - grafana-worker.staging.nubes.stfc.ac.uk
+                  secretName: tls-keypair
```
DavidFair, line 56: For a future PR
Since we have this multiple times, it would be nice to template this into our app of apps - then we can simply pass the `baseDomain`, `secretName`, and `env`

DavidFair (approved):

PR #66, opened 2024-05-08
MAINT: move required charts into deps subdirectory

anish-mudaraddi:
a few charts are required to be installed for argocd to work properly. These charts are now to be stored in deps/ subdirectory to make it cleaner

Since this is just moving charts around and editing paths - testing on staging should be fine to do (prod won't be affected since its tracking another tag)


RJaliliSTFC (approved):

anish-mudaraddi:
I'm going to close this since we're moving to folder-based flows

PR #76, opened 2024-05-30
PROTOTYPE: add argocd applications and appsets

anish-mudaraddi:
add argocd apps/appsets for management cluster (Dev)


clusters/dev/management/argocd-apps/deps.yaml
```diff
@@ -0,0 +1,52 @@
...
+    namespace: argocd
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    repoURL: https://github.com/stfc/cloud-deployed-apps.git
+    targetRevision: HEAD
+    path: charts/dev/deps/argocd
+    helm:
+      valueFiles:
+        - ../../../../clusters/dev/management/argocd-values.yaml
```
DavidFair, line 17: This should just be `argocd-values.yaml` as it's in the same directory looking at the relative path?

clusters/dev/management/argocd-apps/deps.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
```
DavidFair, line 2: Why are we doing an `Application` per thing deployed?
I thought we were doing an ApplicationSet with a list, so we could simply do something like:

```[argocd, certbot, cloud-infra, jupyterhub]```

File: clusters/dev/management/argocd-apps/infra-appset.yaml [hunk 3]
```diff
@@ -0,0 +1,75 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cloud-deployed-infra
```
DavidFair, line 4: This should be in one of the charts, with the ApplicationSet described above pointing to it

clusters/dev/management/argocd-apps/deps.yaml
DavidFair: I think we'd need to undo your change to splitting the charts down to `deps` and `apps` so all of these subcharts are equal

clusters/dev/management/argocd-apps/kustomization.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
DavidFair, line 2: We should drop this completely - just hard-code it to point to the feature branch then we don't have to learn Kustomize for the duration of the feature branch

scripts/deploy-helm-secret.sh
```diff
@@ -0,0 +1,26 @@
...
+SOPS_PRIVATE_KEY_FP="$1"
+
+# Check if the SOPS private key file exists
+if [ -z "$SOPS_PRIVATE_KEY_FP" ]; then
+  echo "SOPS private key file not found."
+  exit 1
+fi
+
+# Extract the third line from the SOPS private key file
+sed -n '3p' "$SOPS_PRIVATE_KEY_FP" > /tmp/key.txt
```
DavidFair, line 21: We shouldn't be trying to parse the key file as it's fraught with potential errors - just put it in as-is and let AGE handle it

It might be worth checking and erroring if n>1 private keys are there

DavidFair (changes requested):

clusters/dev/management/argocd-apps/deps.yaml
anish-mudaraddi: `argocd-values.yaml` is the cluster-specific values - i.e. the DNS name for argocd for the cluster
the default `values.yaml` for our chart gets picked up by default

clusters/dev/management/argocd-apps/deps.yaml
anish-mudaraddi: This is just the dependencies for argocd and our gitops - I separated them out so users who want to use our charts but not argocd wouldn't be confused

clusters/dev/management/argocd-apps/infra-appset.yaml
anish-mudaraddi: in one of the charts? 
I'm confused - I thought you wanted a separate definition per cluster

clusters/dev/management/argocd-apps/kustomization.yaml
anish-mudaraddi: how do I make an Application look at hardcoded paths - I thought it needs to point to either a kustomize.yaml or a chart

scripts/deploy-helm-secret.sh
anish-mudaraddi: it'll just pick up the first private key in the list. 

I want to keep this since I use it -  if other people run into issues with it then they can fix it

clusters/dev/management/argocd-apps/infra-appset.yaml
anish-mudaraddi: I guess I can make a Chart for infra AND argo dependencies?

DavidFair (approved):

PR #84, opened 2024-06-06
DOC: update docs for folder_flows

anish-mudaraddi:
update docs for folder-based flows


docs/charts.md
```diff
@@ -34,10 +38,22 @@ Note: Applications can be held back from a full release if required, e.g. if a p
...
-- Create a new directory under `charts/dev` (or similar) for your new chart
+
+- Create a new directory under `charts/dev` (or similar) for your new chart - you can also choose to deploy a new cluster for testing - see [Deploying a new cluster](clusters.md)
+
 - Simply use `helm install` to install the chart on a test cluster to see if it works, iterating on the chart until it does.
-- Note any environment specific values that cannot ever be shared between environments, e.g. domain name, monitoring....etc.
-- Remove these values and add them into the specific cluster's values file as described in [TODO]
-- Follow the instructions for adding your application to the appset as described in [TODO]
+
+- Note any environment specific values that cannot ever be shared between environments, e.g. domain name, monitoring....etc.
```
DavidFair, line 49: Nit: Ensure there are no environment specific....

And I'd also add a reason why:

"If you have env specific values we cannot promote by simply copying the charts. As a rule of thumb, if you need to consider a value per cluster (e.g. domain name) place it into the `clusters` dir"

docs/clusters.md
```diff
@@ -0,0 +1,164 @@
...
+  addons:
+    ingress:
+      enabled: true
+      nginx:
+        release:
+          values:
+            controller:
+              service:
+                loadBalancerIP: "130.xxx.yyy.zzz"
+
```
DavidFair, line 44: We should turn this into a template yaml file - e.g. at `charts/templates/infra-template.yaml`

That way we can run the linters and easily `cp charts/templates/infra-template.yaml charts/$env/infra-template.yaml`

docs/clusters.md
```diff
@@ -0,0 +1,164 @@
...
+In this file you will define age public keys for those who can decrypt/encrypt secrets belonging to that cluster
+NOTE: Make sure that the management cluster for the same environment can decrypt those secrets. Make sure that the management cluster's public key is added to `.sops.yaml`
+
+7. create file `api-server-fip.yaml` with the following info:
+
+```
+openstack-cluster
+    apiServer:
+        floatingIP: 130.xxx.yyy.zzz
+```
```
DavidFair, line 61: ^
Ditto about having a `secrets/templates` dir

docs/clusters.md
```diff
@@ -0,0 +1,164 @@
...
+
+5. create a new folder under `secrets/<environment>/<cluster-name>`
+
+6. create a new `.sops.yaml` file or copy one from another cluster from the same environment 
+See [Secrets](secrets.md) for more information
+
+In this file you will define age public keys for those who can decrypt/encrypt secrets belonging to that cluster
+NOTE: Make sure that the management cluster for the same environment can decrypt those secrets. Make sure that the management cluster's public key is added to `.sops.yaml`
+
+7. create file `api-server-fip.yaml` with the following info:
```
DavidFair, line 55: This should steer users to use the tooling directly rather than plain-text -> encrypted
It reduces the chances of it being committed plain by accident:

```
7. create the file using `sops api-server-fip.yaml`
```

docs/clusters.md
```diff
@@ -0,0 +1,164 @@
...
+```
+openstack-cluster
+    apiServer:
+        floatingIP: 130.xxx.yyy.zzz
+```
+
+This file contains the floating ip in which Kubernetes API server can be accessed
+MAKE SURE TO ENCRYPT THIS FILE BEFORE COMMITTING - see [Secrets](secrets.md)
+
+8. create file `app-creds.yaml` with the following info:
```
DavidFair, line 66: Then this warning can go away - since `sops` never lets the repo see plain text
And the secrets dir implies this too

docs/clusters.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+# Deploying a new environment
+
+## Prerequisites: 
+1. Create an application credential for your new cluster on your project
+2. Ensure enough quota for the cluster (RAM, CPU, instances etc)
+3. Provision a floating ip on your project for kubernetes API Server access
+4. (Optional) Provision a second floating IP for nginx ingress controller
+5. Create a self-managed cluster called `management` - see https://stfc.atlassian.net/wiki/spaces/CLOUDKB/pages/211878034/Cluster+API+Setup
```
DavidFair, line 110: This must use the same name as your final cluster (since CAPI cannot rename a cluster).
The environment name is not included, i.e. `dev/management`  becomes `management-cluster`

docs/deploying-apps.md
```diff
@@ -0,0 +1,54 @@
+# Deploying a new app onto a cluster
+
+## Prerequisites 
+To deploy a new app onto the cluster - make sure a chart exists for it. 
+A chart must be defined in `charts/<environment>/<chart-name>` 
+    - where `<environment>` is the environment of the cluster you want to deploy the app to
+    - see [charts](charts.md) for more info
+
```
DavidFair, line 8: Understanding AppSet `<link to ArgoCD docs>`

docs/infra-setup.md
```diff
@@ -0,0 +1,117 @@
...
+
+```
+openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout privateKey.key -out certificate.crt
+```
+
+Then to create a secret from a cert - the secret name is for Prometheus, Alertmanager and Grafana is expected to be `tls-keypair` by default:
+
+```
+kubectl create secret tls tls-keypair --cert certificate.crt --key privateKey.key -n monitoring-system
+```
```
DavidFair, line 29: Can we remove this and steer them towards cert-bot
The problem is this differs from dev -> staging -> prod

Using cert-bot means that if you get a self-signed issuer working on dev, it will work with staging and prod automatically too

docs/infra-setup.md
```diff
@@ -0,0 +1,117 @@
...
+                  - hosts: 
+                    - <grafana-hostname>.nubes.stfc.ac.uk
+                    secretName: tls-keypair
+``` 
+
+**NOTE: make sure you follow the post-deployment steps too to configure sending emails**
+
+## Post-deployment
+
+NOTE: this will likely be deprecated for using secrets
```
DavidFair, line 92: Yeah lets add an issue to replace this with secrets directly - but it's fine doc wise at the moment

docs/pics/argocd-app-diagram.png
DavidFair, line 1: Could we use the AppSet names
If they're specific to the cluster name we should probably change that to something like `deployed-apps` and `deployed-infra` to keep it generic

docs/pics/promotion-flowchart.png
DavidFair, line 1: Nit - this should have the starting box with a condition e.g.

> You have a value that needs to go somewhere

docs/secrets.md
```diff
@@ -39,3 +39,10 @@ ArgoCD needs access to the private key to decrypt the secrets. This is done by a
 ```
 kubectl -n argocd create secret generic helm-secrets-private-keys --from-file=key.txt=age-key.txt
 ```
+
+
+## VSCode integration
+
+An extension exists for SOPS for vscode that will automatically decrypt/encrypt files. 
+
+You can find it here: https://github.com/signageos/vscode-sops
```
DavidFair, line 48: Just to check - does this decrypt to plain-text alongside the current file?
If it's decrypts outside of the repo I'm happy to recommend it. If it decrypts alongside the existing secret I'm not as it's very easy to do something like

```
code secret.yaml
# Use the extension to change the secret
# Open a new terminal without closing the currently open secret
git add .   # Now our plain text file is staged 
git commit  # Now we've committed and potentially leaked the secret
```

docs/clusters.md
anish-mudaraddi: future PR

docs/clusters.md
anish-mudaraddi: future PR

docs/clusters.md
anish-mudaraddi: much rather get this feature branch closed first

docs/infra-setup.md
anish-mudaraddi: Whilst I agree - I know that this works and I want to test the cert-bot changes before updating the docs

docs/pics/argocd-app-diagram.png
anish-mudaraddi: only management cluster has deployed-infra so it makes sense to call it `management-apps`

docs/secrets.md
anish-mudaraddi: it does but I've put the file it creates in the .gitignore

docs/clusters.md
anish-mudaraddi: hence why I told them to create a cluster called `management` because that's the name of the management cluster across all environments

DavidFair (approved):

docs/deploying-apps.md
```diff
@@ -0,0 +1,57 @@
...
+          valuesFile: ../../../clusters/<environment>/<cluster-name>/<cluster-specific-values-file>.yaml
+
+```
+
+NOTE: IGNORE THE EXISTING ELEMENTS
+
+`name`: name of ArgoCD-Application
+`chartName`: name of chart - matching directory name should be in `charts/<environment>/`
+`valuesFile`: path to cluster-specific values file for chart (relative to chart location in repo)
+    - NOTE: this `valuesFile` parameter is NON-OPTIONAL - you must provide a file, even if its blank
```
RWhiting78, line 44:
```suggestion
`name`: name of ArgoCD-Application

`chartName`: name of chart - matching directory name should be in `charts/<environment>/`

`valuesFile`: path to cluster-specific values file for chart (relative to chart location in repo)

    - NOTE: this `valuesFile` parameter is NON-OPTIONAL - you must provide a file, even if its blank
```

RWhiting78 (changes requested):

RWhiting78 (approved):

PR #88, opened 2024-06-07
MAINT: Rename chart to match values names

DavidFair:
### Description:

Renames the chart to match our the upstream name, openstack-cluster This makes it easier to reason with including when we're working with our cluster specific values files

Runs an auto-formatter over the diff too - the key one in `apps.yaml` is the chart path has changed to point to this too at line 125

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


meoflynn (commented):
LGTM, but I think we need to update the clusters.md docs with the updated chart name as well as it references `capi-infra` ?

anish-mudaraddi (approved):

PR #89, opened 2024-06-07
Fix appset invalid preserve

DavidFair:
### Description:

The template is unable to preserve an application, since it's defined in
the applicationset instead of an application set creating N applications.
    
This means you can't add preserve to the template, since theres no way for Argo to actually do this

This should fix the auto-sync on dev failing to reconcile currently

### Special Notes:

- See second commit for human diff, the first is auto-formatting

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/clusters/dev/management/apps.yaml
+++ b/clusters/dev/management/apps.yaml
@@ -13,7 +13,6 @@ spec:
     targetRevision: folder_flows
     path: clusters/dev/management
   syncPolicy:
-    preserveResourcesOnDeletion: true
     automated:
       prune: false
       selfHeal: true
--- a/clusters/dev/management/apps.yaml
+++ b/clusters/dev/management/apps.yaml
@@ -54,13 +54,6 @@ spec:
   syncPolicy:
     # Don't remove everything if we remove the appset
     preserveResourcesOnDeletion: true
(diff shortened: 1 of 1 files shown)


clusters/dev/management/apps.yaml
```diff
@@ -22,74 +22,72 @@ spec:
...
+        namespace: "{{namespace}}"
 
----
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+          allowEmpty: true
+        syncOptions:
+          - CreateNamespace=true
```
meoflynn, line 89: With this sync policy in the template, is this the default for apps if it hasn't been defined elsewhere for a specific app in the appset? (e.g. management apps have the sync policy defined in the previous section)

DavidFair: Effectively, it sets all the applications that aren't infra related to automatically prune yes
Currently, there's no way to override this - the rationale being a `git revert` should undo everything in one as it should all be in the repository rather than manually `kubectl apply ...` or similar

meoflynn (approved):

anish-mudaraddi (approved):

PR #95, opened 2024-06-13
FEAT: use folder-based flows

anish-mudaraddi:
### Description:
This PR changes the way we use argocd to use folder-based flows
We've been successfully testing this flow on this feature branch so it is ready to go

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


docs/pics/diagram.drawio
meoflynn, line 1: Could we update the diagram to be a PNG file instead of `.drawio`

charts/dev/capi-infra/values.yaml
```diff
@@ -1,4 +1,11 @@
 openstack-cluster:
+  kubernetesVersion: "1.28.7"
+  machineImage: "capi-ubuntu-2004-kube-v1.28.7-2024-03-01"
+
+  # The PEM-encoded CA certificate for openstack.stfc.ac.uk
+  # this expires 2023-12-05T23:59:59Z (UTC)
```
meoflynn, line 6: I don't think this is the right expiry date?

charts/staging/capi-infra/values.yaml
```diff
@@ -0,0 +1,195 @@
+openstack-cluster:
+  kubernetesVersion: "1.28.7"
+  machineImage: "capi-ubuntu-2004-kube-v1.28.7-2024-03-01"
+
+  # The PEM-encoded CA certificate for openstack.stfc.ac.uk
+  # this expires 2023-12-05T23:59:59Z (UTC)
```
meoflynn, line 6: This doesn't seem like the right date

meoflynn (changes requested):
Just did a quick initial review

charts/staging/capi-infra/values.yaml
meoflynn: Discussed in-person, certificate is still valid

charts/dev/capi-infra/values.yaml
meoflynn: Discussed in-person, certificate is still valid

docs/pics/diagram.drawio
anish-mudaraddi: this is so you can update the diagrams in https://app.diagrams.net

meoflynn (approved):

DavidFair (approved):

PR #99, opened 2024-07-01
BUG: Use app name implicitly

DavidFair:
### Description:

Use the app name implicitly, rather than explicitly saying the name again for the helm release: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-release-name
This ensures that the cluster name matches the user's expectation when it's installed

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/clusters/staging/management/apps.yaml
+++ b/clusters/staging/management/apps.yaml
@@ -111,7 +111,6 @@ spec:
         targetRevision: main
         path: charts/staging/capi-infra
         helm:
-          releaseName: "{{path.basename}}"
           valueFiles:
             # Bring in values that are specific to this application
             - "../../../{{path}}/{{path.filename}}"


clusters/dev/management/apps.yaml
```diff
@@ -98,7 +98,7 @@ spec:
 
   template:
     metadata:
-      name: "{{path.basename}}-cluster"
+      name: "dev-{{path.basename}}-cluster"
```
anish-mudaraddi, line 101: why change this?

anish-mudaraddi (changes requested):

clusters/dev/management/apps.yaml
DavidFair: To bring it into line with staging - I'll drop it and circle back to it when we re-do dev :+1:

anish-mudaraddi (approved):

PR #107, opened 2024-08-01
ENH: Update victoria-metrics chart in dev + create in staging

RJaliliSTFC:
### Description:

Changes to default values for victoria-metrics
Chart copied to staging charts

### Reviewer

Have you:


Chris-green-stfc (approved):

charts/staging/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+    ingress:
+      enabled: true
+      ingressClassName: nginx
+      hosts:
+        - name: select.vm.example.com
+          path: /
+          port: http
+      tls:
+        - secretName: vm-ingress-tls
```
DavidFair, line 13: Can we use the cert manager issuer `self-signed` to make this easier to deploy

charts/staging/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,39 @@
...
+          hosts:
+            - select.vm.example.com
```
anish-mudaraddi, line 15: following on from @DavidFair heres how you'd do it
```suggestion
    ingress:
      enabled: true
      ingressClassName: nginx
      annotations:
          cert-manager.io/cluster-issuer: "self-signed"
      hosts:
        - name: select.vm.example.com
          path: /
          port: http
```

charts/staging/victoria-metrics/values.yaml
```diff
@@ -0,0 +1,39 @@
...
+      enabled: true
+      ingressClassName: nginx
+      hosts:
+        - name: insert.vm.example.com
+          path: /
+          port: http
+      tls:
+        - secretName: vm-ingress-tls
+          hosts:
+            - insert.vm.example.com
```
anish-mudaraddi, line 30: same here 
```suggestion
    ingress:
      enabled: true
      ingressClassName: nginx
      annotations:
          cert-manager.io/cluster-issuer: "self-signed"
      hosts:
        - name: insert.vm.example.com
          path: /
          port: http
```

charts/dev/victoria-metrics/values.yaml
```diff
@@ -35,7 +35,5 @@ victoria-metrics-cluster:
 
     persistentVolume:
       enabled: true
-      existingClaim: ""
-
-      size: 10Gi
-      storageClass: "longhorn"
+      size: 1Gi
+      storageClass: "csi-manila-cephfs"
```
anish-mudaraddi, line 39: how are you going to deploy manila? It will have to be available on the cluster beforehand
We'll have to add it in as a chart and an app for argocd to manage

anish-mudaraddi: I've deployed manila on dev - you can try and use it if you want

RJaliliSTFC:
Needs to be worked on time permitting

PR #118, opened 2024-08-09
PROTOTYPE: Enable monitoring on dev

anish-mudaraddi:
We had to turn off storage since cinder is broken on dev. 
We've disabled sending alerts for this cluster

Places shared config for monitoring in chart values ready for promotion


anish-mudaraddi:
solves - https://github.com/stfc/cloud-deployed-apps/pull/118

PR #138, opened 2024-08-23
update capi image to 1.28.12

anish-mudaraddi:
update capi image on dev to test


charts/dev/capi-infra/values.yaml
```diff
@@ -1,6 +1,6 @@
 openstack-cluster:
   kubernetesVersion: "1.28.7"
```
meoflynn, line 2: You will need to update the version here to 1.28.12 as well

meoflynn (changes requested):

meoflynn (approved):

PR #143, opened 2024-09-09
DEPLOY: Move prod to use folder flows

gmatthews20:


anish-mudaraddi (approved):
Worked on this together and I'm happy with the changes - we're just mirroring staging here

clusters/prod/management/infra-values.yaml
```diff
@@ -0,0 +1,44 @@
+openstack-cluster:
+  controlPlane:
+    machineCount: 3
```
DavidFair, line 3: This should be 5, actually we should drop this and use the chart default unless we need to tailor it

clusters/prod/management/infra-values.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+  nodeGroups:
+    - name: default-md-0
+      machineCount: 2
+
+  nodeGroupDefaults:
+    machineFlavor: l3.nano
```
DavidFair, line 10: ^

clusters/prod/worker/infra-values.yaml
```diff
@@ -0,0 +1,44 @@
+openstack-cluster:
+  controlPlane:
+    machineCount: 3
+
+  nodeGroups:
+    - name: default-md-0
+      machineCount: 2
+
+  nodeGroupDefaults:
+    machineFlavor: l3.nano
```
DavidFair, line 10: ^

secrets/prod/management/.sops.yaml
```diff
@@ -0,0 +1,8 @@
+creation_rules:
+  - unencrypted_regex: "^(apiVersion|metadata|kind|type)$"
+    key_groups:
+      - age:
+          # prod-rotate-key
+          - age1cq92796c46d06s43t079xc89exe4vd52rh30c9mcafmne62dxyhqrupl4l
+          # staging-argo-key
+          - age1mqqfadar4vz8yfsvfd540nkhwjvxuzmf3hgr95ftp7d0dqnm7ycsfue3kv
```
DavidFair, line 8: Is this the staging key? It should be a dedicated prod key

DavidFair (changes requested):

secrets/prod/management/.sops.yaml
gmatthews20: This is the prod key, I have updates the comment

anish-mudaraddi (approved):

DavidFair (approved):

PR #145, opened 2024-09-10
Promote capi infra, capi addon provider and cert manager charts to staging

gmatthews20:
Promote capi infra, capi addon provider and cert manager charts to staging


anish-mudaraddi (approved):
happy with these changes - monitoring is not currently enabled on staging but it should be - in a follow up PR

meoflynn (approved):

PR #160, opened 2024-09-27
New workflow to detect whether files within charts directory match

DaveW-STFC:
The included workflow file will find any changes made under the Charts folder for the staging and prod environments.
It will then pass each of these to the bash script for comparison and return whether they match with the previous environment(s).
Please have a look over and give any thoughts on the process, code etc.

### DIFF
--- a/.github/workflows/pr_copied_files.yaml
+++ b/.github/workflows/pr_copied_files.yaml
@@ -1,0 +1,102 @@
+name: CI Copied Files Check
+
+on:
+  push:
+    branches:
+      - check_pr_copies
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  changed_files:
+    runs-on: ubuntu-latest
+    name: Test changed-files
+    outputs:
+      staging_result: ${{ steps.staging-check.outputs.result }}
+      prod_result: ${{ steps.prod-check.outputs.result }}
+      any_changed: ${{ steps.changed-files.outputs.any_changed }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: check_pr_copies
+      - name: Get changed files in specific repo folders
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          files_yaml: |
+            staging:
+              - charts/staging/**
+            prod:
+              - charts/prod/**
+          #  clusters/staging/**
+          #  clusters/prod/**
+          #  secrets/staging/**
+          #  secrets/prod**
+
+      - name: Run step if any file(s) in staging env change
+        id: staging-check
+        if: steps.changed-files.outputs.staging_any_changed == 'true'
+        env:
+          STAGING_ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.staging_all_changed_files }}
+        run: |
+          echo "One or more files in the staging folder has changed."
+          echo "Files that have changed: $STAGING_ALL_CHANGED_FILES"
+          echo "result=true" >> "$GITHUB_OUTPUT"
+          for file in ${{ steps.changed-files.outputs.staging_all_changed_files }}; do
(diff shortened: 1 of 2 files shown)


anish-mudaraddi:
@DaveW-STFC - could you squash all these test commits down into 1 or 2

.github/workflows/pr_copied_files.yaml
```diff
@@ -0,0 +1,64 @@
+name: CI Copied Files Check
```
anish-mudaraddi, line 1: just a suggestion - but why don't you split this entire workflow into 2 parts:
- `check_staging_integrity` - one for comparing staging file changes with dev
- `check_prod_integrity` - one for comparing prod file changes with staging

This allows us to more clearly see where the issue is at a glance.

It also allows you to modify your script - instead of hardcoding the values `staging` and `dev` for `stagingEnv` and `devEnv` you can pass it as an argument 
- `/pr_check_copies.sh "$file" <env-to-compare-against>` - makes it more versatile in case we add more environments for instance

scripts/pr_check_copies.sh
```diff
@@ -0,0 +1,39 @@
...
+
+    # Check the number of arguments passed to the function for comparison
+    if [ "$#" -eq 2 ]; then
+        if cmp -s "$file1" "$file2" ; then
+            echo $((0))
+        else
+            echo $((1))
+        fi
+    else
+        if cmp -s "$file1" "$file2" && cmp -s "$file1" "$file3"; then
```
anish-mudaraddi, line 23: will this work if the file doesn't exist in the environment your comparing against? I suspect we'll need a test case for that to output nicely

anish-mudaraddi (changes requested):
good start - a few things to consider

.github/workflows/pr_copied_files.yaml
DaveW-STFC: Will prod only be compared to staging or staging + dev?

DavidFair: It should be staging, as we shouldn't promote to dev -> prod without good reason 

If this job fails it indicates we should use a 2 person review, so I'm happy to just do `dev -> staging`, `staging -> prod`
If staging doesn't match someone is doing a `dev -> prod` which would fail with yours (which is what we want)

DaveW-STFC: I will need to update this then as i've set it to check any promotes to Prod match both staging and dev as it was mentioned in the task. There is a chance dev would not match as other changes could have gone into dev since promotion to staging. Removing the comparison to both makes it much simpler on the script side.
@anish-mudaraddi thoughts?

DaveW-STFC: I've simplified the script to account for this so things are much simpler.

.github/workflows/pr_copied_files.yaml
```diff
@@ -0,0 +1,95 @@
...
+            res=$( ${{ github.workspace }}/.github/workflows/pr_check_copies.sh "$file" "prod" "staging" )
+            echo $res
+            if [ "$res" -eq 1 ]; then
+              echo "result=false" >> "$GITHUB_OUTPUT"
+              break;
+            fi
+          done
+
+  comment_pr:
+    needs: changed_files
```
DavidFair, line 68: I'm not sure if this works with needs and the previous job fails (as it's an unusual workflow), have we tested it where something has changed at all? (e.g. adding a random file in this PR)

IIRC, `needs:` will abort the job if the previous step fails since it needs that output
However we're using it where we need the previous job to fail to run this one

DaveW-STFC: We want the previous job to succeed as we are comparing the files to check if they are a direct copy from previous environment. If they are then we want to add a comment to state so. If they don't match then we do not comment and the review should be carried out as there are changes to check.

PR #173, opened 2024-10-03
BUG: use crds.enable=true

anish-mudaraddi:
### Description:

InstallCRDs flag is deprecated - use `crds.enable and crds.keep` instead

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/charts/staging/cert-manager/values.yaml
+++ b/charts/staging/cert-manager/values.yaml
@@ -9,4 +9,6 @@ cert-manager:
   le-prod:
     enabled: false
 
-  installCRDs: true
+  crds:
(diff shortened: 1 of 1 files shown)


charts/staging/cert-manager/values.yaml
```diff
@@ -9,4 +9,6 @@ cert-manager:
   le-prod:
     enabled: false
 
-  installCRDs: true
+crds:
```
RJaliliSTFC, line 12: I think indentation is wrong

RJaliliSTFC (approved):

PR #175, opened 2024-10-03
Update deploy.sh

RJaliliSTFC:
Add a temp fix to get cert-manager crds

### Description:

Added two lines to deploy.sh to get a repo and install CRDs for cert-manager

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -34,7 +34,7 @@ fi
 
 # Installing dependencies for cert-manager (temp fix)
 helm repo add jetstack https://charts.jetstack.io --force-update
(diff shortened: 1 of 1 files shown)


anish-mudaraddi (approved):
need a second reviewer for this

anish-mudaraddi:
We need to install cert-manager because if we're deploying a child cluster - cert-manager crds won't be installed - and argoCD requires them in order to start. Best way to install the CRDs is just to install the helm chart - we then get argoCD to manage the helm chart after it's spin up using gitops 

Moving forwards - I think we should make cert-manager a dependency of argoCD chart and remove cert-manager chart altogether

PR #181, opened 2024-10-07
MAINT: remove group requirement for galaxy iris-iam login

anish-mudaraddi:
needed so PSDI can test this without adding them to stfc-cloud-dev group

### DIFF
--- a/clusters/dev/galaxy/materials-galaxy-values.yaml
+++ b/clusters/dev/galaxy/materials-galaxy-values.yaml
@@ -1,7 +1,7 @@
 oauth2-proxy:
   extraArgs:
     # limit access to stfc-cloud-dev group
-    allowed-group: "stfc-cloud-dev"
+    # allowed-group: "stfc-cloud-dev"
     redirect-url: "https://galaxy.dev.nubes.stfc.ac.uk/oauth2/callback"
 
 galaxy:


clusters/dev/galaxy/materials-galaxy-values.yaml
```diff
@@ -1,7 +1,7 @@
 oauth2-proxy:
   extraArgs:
     # limit access to stfc-cloud-dev group
-    allowed-group: "stfc-cloud-dev"
+    # allowed-group: "stfc-cloud-dev"
```
DavidFair, line 4: Won't this allow anyone in from IRIS IAM?
They should have a PSDI group we can add, and (if possible) we should retain dev access on the dev cluster

DavidFair (changes requested):

clusters/dev/galaxy/materials-galaxy-values.yaml
meoflynn: I also agree, is there an IAM group we can add for PSDI?

anish-mudaraddi: there isn't an IAM group  - and since the url has already been shared to PSDI and they intend to test things out -  I'm inclined to remove allowed-group. I can instead limit it to the emails of PSDI members if security is an issue

DavidFair: If they're aware of this limitation I'm happy to let it roll, as it's their call on balance between getting going and getting groups in

anish-mudaraddi:
I'll reopen this once PSDI reply with a replacement group or authorized email addresses

PR #183, opened 2024-10-07
BUG: add env label to prod cluster alerts

anish-mudaraddi:
env label must be prod so we know where the alerts are coming from

### DIFF
--- a/clusters/prod/management/infra-values.yaml
+++ b/clusters/prod/management/infra-values.yaml
@@ -15,7 +15,8 @@ openstack-cluster:
           values:
             defaultRules:
               additionalRuleLabels:
-                cluster: prod-management
+                cluster: management
+                env: prod
             prometheus:
               ingress:
                 hosts:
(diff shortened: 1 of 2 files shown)


clusters/prod/management/infra-values.yaml
```diff
@@ -15,7 +15,8 @@ openstack-cluster:
           values:
             defaultRules:
               additionalRuleLabels:
-                cluster: prod-management
+                cluster: management
+                env: prod
```
DavidFair, line 19: Do we need to apply the same changes to anything in `dev` or staging?

anish-mudaraddi: this is just a bug for prod - see #154 for better way of setting alerting labels (using `external_labels`)

anish-mudaraddi: I'll promote #154 changes to prod after testing on staging for a bit longer

PR #187, opened 2024-10-08
BUG: disable galaxy

anish-mudaraddi:
there's a bug with galaxy with argocd where duplicate setup jobs keep getting instantiated - looking to disable galaxy and test manually deploying galaxy to rule out if this is an application issue

Also deleting capi-addon-provider - this is a child cluster so I don't think its necessary


DavidFair (approved):

DavidFair (approved):

DavidFair:
Since these changes are going to be on dev in a dedicated branch we'll go for a single review as it's just changing the deployment
Any fixes we want to get back into `main` we'll do a two person review again

PR #189, opened 2024-10-09
MAINT: add Opensearch

anish-mudaraddi:
Add opensearch chart to dev. This uses opensearch-k8s-operator. 
This also allows iris-iam config setup

### DIFF
--- a/charts/dev/opensearch/templates/opensearch-cluster.yaml
+++ b/charts/dev/opensearch/templates/opensearch-cluster.yaml
@@ -2,7 +2,7 @@ apiVersion: opensearch.opster.io/v1
 kind: OpenSearchCluster
 metadata:
   name: {{ .Values.clusterName }}
-  namespace: {{ .Release.Namespace | default .Values.namespace }}
+  namespace: {{ .Release.Namespace }}
   annotations:
     "helm.sh/hook": post-install,post-upgrade
 spec:
--- a/charts/dev/opensearch/templates/opensearch-cluster.yaml
+++ b/charts/dev/opensearch/templates/opensearch-cluster.yaml
@@ -1,0 +1,81 @@
+apiVersion: opensearch.opster.io/v1
+kind: OpenSearchCluster
+metadata:
+  name: {{ .Values.clusterName }}
+  namespace: {{ .Release.Namespace | default .Values.namespace }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+spec:
+  security:
+    config:
+      adminCredentialsSecret:
+        name: {{ .Values.adminCredentialsSecret }}
+      securityConfigSecret:
+        name: {{ .Values.securityConfigSecret }}
+    tls:
+      transport:
+        generate: true
+        perNode: true
+      http:
+        generate: true
+  general:
+    serviceName: {{ .Values.clusterName }}
+    setVMMaxMapCount: true
+    version: {{ .Values.version }}
+    httpPort: {{ .Values.httpPort | default 9200 }}
+    monitoring: {{ toYaml .Values.monitoring | nindent 6 }}
+    pluginsList: {{ toYaml .Values.pluginsList | nindent 6 }}
+  confMgmt:
+    smartScaler: {{ .Values.useSmartScaler | default false }}    
+  dashboards:
+    tls:
+      enable: true
+      generate: true
+    enable: {{ .Values.dashboards.enabled | default true }}
+    version: {{ .Values.version }}
+    replicas: {{ .Values.dashboards.replicas | default 1 }}
+    opensearchCredentialsSecret:
+      name: {{ .Values.adminCredentialsSecret }}
+    env:
+      - name: IAM_CLIENT_ID
+        valueFrom:
+          secretKeyRef:
+            name: iris-iam-credentials-secret
+            key: client-id
+      - name: IAM_CLIENT_SECRET
+        valueFrom:
+          secretKeyRef:
+            name: iris-iam-credentials-secret
+            key: client-secret
+    additionalConfig:
+      opensearch_security.multitenancy.enabled: "true"
+      opensearch_security.cookie.secure: "true"
+      opensearch_security.auth.type: | 
+        ["basicauth", "openid"]
+      opensearch_security.auth.multiple_auth_enabled: "true"
+      opensearch_security.openid.connect_url: "https://iris-iam.stfc.ac.uk/.well-known/openid-configuration"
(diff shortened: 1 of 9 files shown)


charts/dev/opensearch/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+    nginx.ingress.kubernetes.io/proxy-body-size: "0"
+  hosts:
+    - host: nodes.demo.io
+      paths:
+        - path: /
+          pathType: Prefix
+  tls:
+    - secretName: opensearch-tls
+      hosts:
+        - nodes.demo.io
```
DavidFair, line 72: We don't own this domain, so it should either be an `example.com` based domain or one we own

docs/apps/opensearch.md
```diff
@@ -0,0 +1,206 @@
...
+
+You'll need to setup users, roles and rolebindings here in the appropriate files. We cannot use CRDs for defining these whilst also setting up oidc 
+
+You will also need to create a Kubernetes Secret containing admin credentials (encoded in base64) PLEASE CHANGE THE PASSWORD FROM ADMIN:
+
+```bash
+kubectl create secret generic opensearch-admin-credentials
+--from-literal=username='admin' 
+--from-literal=password='admin'
+-n opensearch-system
```
DavidFair, line 106: This should use SOPS to provide a secret at rest

DavidFair (changes requested):

docs/apps/opensearch.md
anish-mudaraddi: we'll need to think more about our appset design

RJaliliSTFC (approved):

docs/apps/opensearch.md
anish-mudaraddi: I'll make a follow up PR for this

charts/dev/opensearch/templates/opensearch-nodes-ingress.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.ingress.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ .Values.clusterName }}-nodes-ingress
```
DavidFair, line 5: Nit: `worker-nodes-ingress` isn't a great name, could we have `worker-opensearch-ingress` ?

charts/dev/opensearch/templates/opensearch-nodes-ingress.yaml
```diff
@@ -0,0 +1,38 @@
...
+  namespace: {{ .Release.Namespace | default "opensearch" }}
```
DavidFair, line 6: This should just use the `Release.Namespace` rather than falling back to defaults which we have to spread throughout

charts/dev/opensearch/templates/opensearch-cluster.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: opensearch.opster.io/v1
+kind: OpenSearchCluster
+metadata:
+  name: {{ .Values.clusterName }}
+  namespace: {{ .Release.Namespace | default .Values.namespace }}
```
DavidFair, line 5: In some places we use `Release.Namespace | default .Values.namespace`

Reading around other projects just use `Release.Namespace` so I think we can drop the values one then let ArgoCD + Helm get the NS correct (since this is overwritten by argo `Application` CRD)

charts/dev/opensearch/templates/opensearch-dashboards-ingress.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.dashboards.ingress.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ .Values.clusterName }}-dashboards-ingress
+  namespace: {{ .Release.Namespace | default "opensearch-system" }}
```
DavidFair, line 6: Ditto

charts/dev/opensearch/values.yaml
```diff
@@ -0,0 +1,151 @@
+opensearch-operator:
+  installCRDs: true
+
+clusterName: dev-opensearch
```
DavidFair, line 4: This shouldn't be in this level of chart - since we can't `cp -r dev/opensearch staging/opensearch` with this in

Either we should leave it blank and fail the deployment if it's missing, or have something generic like `opensearch-cluster`

charts/dev/opensearch/values.yaml
```diff
@@ -0,0 +1,151 @@
+opensearch-operator:
+  installCRDs: true
+
+clusterName: dev-opensearch 
+namespace: opensearch-system
+
+# opensearch and opensearch dashboards version
+version: 2.17.0
```
DavidFair, line 8: Could we add a URL to where the versions are kept for future readers - since I assumed this would be set in the upstream chart and we implicitly get it from the version we've got pinned?

charts/dev/opensearch/values.yaml
```diff
@@ -0,0 +1,151 @@
...
+# define new internal users here
+#https://opensearch.org/docs/latest/security/configuration/yaml/#internal_usersyml
+# recommended to setup reserved users here - those that can't be changed using the REST API or OpenSearch Dashboards
+users:
+#  fluentbit:
+#    hash: THIS SHOULD BE A SECRET - set it in using SOPS
+#    reserved: true
+#    backend_roles:
+#    - "kibana_user"
+#    description: "Fluentbit user for sending logs"
```
DavidFair, line 116: Do we have a secret template for this like the OIDC, or is that planned for the future?

charts/dev/opensearch/templates/opensearch-nodes-ingress.yaml
anish-mudaraddi: opensearch ingress is being used for opensearch dashboards - matches the dns opensearch.nubes.stfc.ac.uk
nodes is being used for access to the nodes - i.e. the opensearch api - nodes.nubes.stfc.ac.uk

charts/dev/opensearch/values.yaml
anish-mudaraddi: if you need to deploy a user - you should add the hash of that user into the secret - this is documented

This actually needs updating away from using crds

charts/dev/opensearch/values.yaml
anish-mudaraddi: I will delete it to force people to provide it

DavidFair (approved):

PR #190, opened 2024-10-11
Creating test cluster using Argo/CAPI

DaveW-STFC:
Configuration files to create a new test cluster under the dev env within my own scratch environment. Hopefully!


DavidFair:
Nit: can we rename it to something like `DaveW` instead of your fed ID so we don't have to map back?

anish-mudaraddi:
just be aware your cluster name will be `dev-<whatever name you give it>-cluster` - so you might want to keep it short :)

anish-mudaraddi:
you'll want to rotate the `secrets/dev/smtp-smarthost.yaml` secret to add your age key as well

RJaliliSTFC (approved):

RJaliliSTFC:
> you'll want to rotate the `secrets/dev/smtp-smarthost.yaml` secret to add your age key as well

Just need to look at this @DaveW-STFC

RJaliliSTFC (approved):

PR #192, opened 2024-10-15
MAINT: simplify alertmanager routing

anish-mudaraddi:
Simplify alertmanager routing - we only need one route - for emailing.

Just Inhibit the watchdog and infoinhibitor alerts so they don't send notifications

add a time range where emails are active - during office hours

### DIFF
--- a/charts/dev/capi-infra/values.yaml
+++ b/charts/dev/capi-infra/values.yaml
@@ -187,42 +187,43 @@ openstack-cluster:
                   smtp_require_tls: false
 
                 route:
-                  receiver: 'null'
-                  routes:
-                    - receiver: 'null'
-                      matchers:
-                      - alertname = "Watchdog"
-                    - receiver: default-receiver
-                      group_wait: 30m
-                      group_interval: 30m
-                      repeat_interval: 2d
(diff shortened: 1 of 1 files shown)


charts/dev/capi-infra/values.yaml
```diff
@@ -187,42 +187,43 @@ openstack-cluster:
...
-                      - 'alertname = InfoInhibitor'
+                    - alertname: Watchdog
+                    - alertname: InfoInhibitor
+
+                time_intervals:
+                  - name: outofoffice
+                    time_intervals:
+                      - times:
+                        - start_time: 07:00
+                          end_time: 18:00
```
meoflynn, line 219: Would this mean we don't get alerts if this time interval is out of office?

meoflynn (commented):
I just have one question about the time intervals

charts/dev/capi-infra/values.yaml
anish-mudaraddi: sorry - this is badly named - should be `officehours` instead 
we'll only get alerts when we're in office

meoflynn (approved):

PR #202, opened 2024-10-31
Update ingress IPs and DNS entries for prod clusters

DavidFair:
### Description:
Updates the ingress IPs to reflect the wildcard DNS entry Switch the existing services to start using the wildcard subdomains

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/clusters/prod/management/infra-values.yaml
+++ b/clusters/prod/management/infra-values.yaml
@@ -7,7 +7,8 @@ openstack-cluster:
           values:
             controller:
               service:
-                loadBalancerIP: "130.246.81.161"
+                # *.prod-mgmt.nubes.stfc.ac.uk
+                loadBalancerIP: "130.246.80.233"
 
     monitoring:
       kubePrometheusStack:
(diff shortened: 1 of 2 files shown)


clusters/prod/management/infra-values.yaml
```diff
@@ -7,7 +7,8 @@ openstack-cluster:
           values:
             controller:
               service:
-                loadBalancerIP: "130.246.81.161"
+                # *.prod-mgmt.nubes.stfc.ac.uk
+                loadBalancerIP: "130.246.80.233"
```
RJaliliSTFC, line 11: Why are we changing the IP here? This IP has the description 'staging'

clusters/prod/worker/infra-values.yaml
```diff
@@ -20,25 +21,25 @@ openstack-cluster:
...
               ingress:
                 hosts:
-                  - alertmanager-worker.prod.nubes.stfc.ac.uk
+                  - alertmanager.prod-worker.nubes.stfc.ac.uk
                 tls:
                   - hosts:
-                      - alertmanager-worker.prod.nubes.stfc.ac.uk
-                    secretName: tls-keypair
\ No newline at end of file
+                      - alertmanage.prod-worker.nubes.stfc.ac.uk
```
RJaliliSTFC, line 44: Mispelling?

clusters/prod/management/infra-values.yaml
DavidFair: If you `dig prod-mgmt.nubes.stfc.ac.uk` you get this IP? I suspect the description against the FIP is wrong

RJaliliSTFC (approved):

PR #213, opened 2024-11-19
MAINT: make galaxy use secrets

anish-mudaraddi:
Galaxy application now utilises sops secrets to set admin emails and IAM credentials
Added Docs for Galaxy deployment
Also I removed manila csi application as we're not using it


charts/dev/materials-galaxy/secrets-templates.yaml/materials-galaxy.yaml
RJaliliSTFC, line 1: Do you want this directory to be called secrets-templates.yaml?

charts/dev/materials-galaxy/secrets-templates.yaml/materials-galaxy.yaml
```diff
@@ -0,0 +1,35 @@
...
+    clientID: iris-iam-client-id
+    clientSecret: iris-iam-secret
+    # Create a new secret with the following command
+    # openssl rand -base64 32 | head -c 32 | base64
+    cookieSecret: "XXXXXXXXXXXXXXXX"
+  
+  # if you want to setup galaxy where only specific emails have access
+  # USE EITHER THIS OR RESTRICT BY OIDC GROUP 
+  authenticatedEmailsFile:
+    enabled: true
```
RJaliliSTFC, line 14: if this is enabled true by default, will it error if no file is provided?

clusters/dev/galaxy/materials-galaxy-values.yaml
```diff
@@ -1,16 +1,11 @@
 oauth2-proxy:
+
+
   extraArgs:
     # limit access to stfc-cloud-dev group
```
RJaliliSTFC, line 5: Comment is old now?

RJaliliSTFC (changes requested):

charts/dev/materials-galaxy/secrets-templates.yaml/materials-galaxy.yaml
anish-mudaraddi: any better names you can think of? Its just to store "an example secret" which you can copy-paste into an encrypted sops file and change the values

charts/dev/materials-galaxy/secrets-templates.yaml/materials-galaxy.yaml
anish-mudaraddi: This is just a template - so its not being used - but fair point I'll write a comment to set this to false it if you're restricting by OIDC group

charts/dev/materials-galaxy/secrets-templates.yaml/materials-galaxy.yaml
RJaliliSTFC: Comment was mostly for the .yaml part, think it'd be more clear as just "secrets-templates"

anish-mudaraddi: oh yea - typo :sweat_smile:

RJaliliSTFC (approved):

PR #220, opened 2024-11-20
Promotion of victoria metrics, Staging -> Prod

RJaliliSTFC:
### Description:

Copied app charts from staging to prod
Created new cluster files comparable to staging, changed to point to prod values
Creating a running prod version of Victoria-Metrics

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/clusters/prod/worker/apps.yaml
+++ b/clusters/prod/worker/apps.yaml
@@ -91,7 +91,10 @@ spec:
           allowEmpty: true
         syncOptions:
           - CreateNamespace=true
-
+  
+  # Clusterrole and clusterrolebinding get updated after creation by the helm chart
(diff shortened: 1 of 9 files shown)


clusters/prod/worker/apps.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+          allowEmpty: true
+        syncOptions:
+          - CreateNamespace=true
+
+  templatePatch: |
```
DavidFair, line 95: Can we comment why this is required as it's unusual?

RJaliliSTFC: @anish-mudaraddi

DavidFair (approved):

PR #231, opened 2024-12-02
ENH: Add chatops to dev

khalford:
### Description:

Add a Chart for chatops and an app in the dev worker cluster

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


charts/dev/chatops/values.yaml
```diff
@@ -0,0 +1,7 @@
+env: dev
+name: cloud-chatops-deployment
```
anish-mudaraddi, line 2: feel like you don't need this - as I mentioned in previous comments - you can set a hardcoded name - with some prefix like env. 

A usecase for something like this is If you want the user to specify some prefix so they can deploy multiple versions of your chart (with different names but same env) - you can set it here 

I think this is what you're using env for anyway - maybe a better name for this like `namePrefix` or something idk

clusters/dev/worker/apps.yaml
```diff
@@ -59,6 +59,11 @@ spec:
             valuesFile: ../../../clusters/dev/worker/materials-galaxy-values.yaml
             secretsFile: ../../../secrets/dev/worker/apps/materials-galaxy.yaml
 
+          - name: chatops
+            chartName: chatops
+            namespace: chatops
+            secretsFile: ../../../secrets/dev/worker/chatops.yaml
```
anish-mudaraddi, line 65:
```suggestion
            secretsFile: ../../../secrets/dev/worker/apps/chatops.yaml
```

secrets/dev/worker/apps/.sops.yaml
```diff
@@ -18,3 +18,5 @@ creation_rules:
           - age16fufeddr0arrns268526gxethxgkh3g0euf8cn37kuwfmq3h23psutz4q8
           # Ryan Whiting
           - age1a8e4gxw67kp27s3hssfxyem3e8jwaha3huz0sttfngeu60pk5pxqkfpg3d
+          # Kalibh Halford
+          - age1drky6caal0j2x58yzpw9tyflcpdpmcjqy8nss7zfvspszg0xfpdsyzu8s4
```
anish-mudaraddi, line 22: nice - so you've added your key and used it to make the file below - good! Now this should work fine if deployed 

Since you've encrypted your new secrets file with all keys mentioned here (including your own) - we should all be able to decrypt and validate the secrets. 

The only thing here is that you can't access any previously made secrets. To do this - we need to run `sops updatekeys` to re-encrypt the existing files to add your new key. There's a handy script in `scripts/update-secrets <environment> <clustername>` which does this for you

This can only be done by someone with a key that is shared by all files in directory - I will do this and make a commit

anish-mudaraddi (changes requested):
good start! The chart needs a few changes before I approve. I'll put your sops key on all dev clusters in a new PR - you can rebase your changes on-top of that once merged

charts/dev/chatops/templates/config-map.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: {{ .Values.env }}-cloud-chatops-config
+  namespace: cloud-chatops
+data:
+  config.yml: |
+  ----
+  ---\r
+  user-map:\r
+    khalford: U07P058AF51\r
+  \r
```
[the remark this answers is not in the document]
khalford: I made a configMap from a file using `kubectl create configmap cloud-chatops-config --from-file config.yml -n cloud-chatops` then used `kubectl describe configmap cloud-chatops-config -n cloud-chatops` to get the correct format and it gave me those carriage returns. I will remove them if it's not correct.

charts/dev/chatops/secrets-templates/secrets.yaml
```diff
@@ -0,0 +1,4 @@
+secrets: |
+  SLACK_BOT_TOKEN: <>
```
anish-mudaraddi, line 2: you need to change this to match the values you expect in templates

```
secrets:
    slackBotToken: <>
    slackAppToken: <>
    githubToken: <>
```

anish-mudaraddi: also please add comments for what each does - why its needed etc - useful for users

charts/dev/chatops/templates/config-map.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.env }}-cloud-chatops-config
+  namespace: cloud-chatops
+data:
+  config.yml: |
+  ---
+  user-map:
+    khalford: U07P058AF51
```
anish-mudaraddi, line 10: maybe something like 
```
user-map:
    - name: khalford
      id: <>
```
more informative - might need a bit of refactoring in your code

secrets/dev/worker/apps/chatops.yaml
```diff
@@ -0,0 +1,78 @@
+secrets:
+    slackBotToken: ENC[AES256_GCM,data:y69BPLMV+boj7++PSRieCaA3DylCI9c5nrgioge+VvTjzhFfnJePO5IUoBrjTl8Nz1lzEu0LAvD2,iv:30PMrR7K/j/sIKu7AuuTY3G3IotBTFJJd7xPUgOJJzA=,tag:wuWYFSSop2Lp56fUaDBf+A==,type:str]
+    slackAppToekn: ENC[AES256_GCM,data:2gXRilzYyukxZF9dBd055Xg0uhyTIUyvF/AxX54J9r3wpNjzV+Mbozq/CiRazhUvR1wSrDy7j0f8iTgrgX5XV8IeIHASUpQJZqqRk3oq10TJjTlPcpmQSknk5yYUVT23cQ==,iv:vcbiGES6dwfi5nV+ECYKKFOhJea6mShjI4J+S7lSo8o=,tag:0boDpdgh0TOtHEahnE+Cug==,type:str]
```
anish-mudaraddi, line 3: typo

anish-mudaraddi (changes requested):
you also need some docs for this chart

charts/dev/chatops/templates/config-map.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.env }}-cloud-chatops-config
+  namespace: cloud-chatops
+data:
+  config.yml: |
+  ---
+  users:
+    - Kalibh Halford:
```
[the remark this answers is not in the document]
anish-mudaraddi: if you opt for above change to secrets format - namely: 
```
users:
  - name: foo
    githubUser: ""
    slackID: ""
  - name: bar
    githubUser: "somevalue"
    slackID: "anothervalue"
```

you can do something like this 
```
data:
   config.yml: | 
    ---
   users:
      {{- range .Values.secrets.users }}
      - {{ .name }}:
          github_name: {{ .githubName | quote }}
          slack_id: {{ .slackID | quote }}
      {{- end }}
```

charts/dev/chatops/templates/config-map.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - cloud-image-builders
+      - cloud-docs
+      - cloud-openstack-horizon
+      - ansible-harbor
+      - terraform-openstack
+      - cloud-rundeck-jobs
+      - openstack-guide
+      - ansible-jupyter
+      - SCD-Openstack-Utils
+      - check-version-action
```
anish-mudaraddi, line 31: you should really make this part of values.yaml - easier to edit - makes it reusable for others outside of our team

```
repos: {{ toYaml  .Values.repos | nindent 4 }}
```

anish-mudaraddi (changes requested):
you still need docs :)

charts/dev/chatops/templates/cloud-chatops.yaml
anish-mudaraddi, line 1: Does your container have any endpoints that needs to be called/accessed or is it purely making outbound API calls to slack/github?
If you want to accept inbound connections you need to make a service component (and ingress resource) - not sure if this applies to your stuff

charts/dev/chatops/secrets-templates/secrets.yaml
anish-mudaraddi: you've not resolved this - you need to change these to match what your helm templates are expecting

anish-mudaraddi (changes requested):

charts/dev/chatops/secrets-templates/secrets.yaml
khalford: Yep missed that

charts/dev/chatops/templates/cloud-chatops.yaml
khalford: No endpoints only outbound.

anish-mudaraddi (approved):

gmatthews20 (approved):

PR #242, opened 2024-12-12
Update to v5

khalford:
### Description:

Update to version 5 of the chatops image. This gives us access to the `/find-host` comand. For this we need to add the node IP to the environment variables and add the pull requests channel to the config.

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/charts/dev/chatops/templates/cloud-chatops.yaml
+++ b/charts/dev/chatops/templates/cloud-chatops.yaml
@@ -21,6 +21,11 @@ spec:
       containers:
       - name: {{ .Values.app }}
         image: {{ .Values.image.repository }}:{{ .Values.image.tag}}
+        env:
+        - name: HOST_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
         volumeMounts:
         - name: cloud-chatops-config
           mountPath: /usr/src/app/cloud_chatops/config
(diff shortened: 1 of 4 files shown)


charts/dev/chatops/values.yaml
```diff
@@ -23,3 +23,4 @@ repos:
     - ansible-jupyter
     - SCD-Openstack-Utils
     - check-version-action
+channel: C03RT2F6WHZ
```
anish-mudaraddi, line 26: Should this be a secret?

charts/dev/chatops/templates/cloud-chatops.yaml
```diff
@@ -21,6 +21,11 @@ spec:
       containers:
       - name: {{ .Values.app }}
         image: {{ .Values.image.repository }}:{{ .Values.image.tag}}
+        env:
+        - name: HOST_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
```
anish-mudaraddi, line 28: what's this for?

anish-mudaraddi (changes requested):
have some questions

charts/dev/chatops/values.yaml
khalford: Not really, it is just a channel ID. It is only accessible from inside our workspace so I see no issue with it being public.

charts/dev/chatops/templates/cloud-chatops.yaml
khalford: This is for pod to access its node IP address. This is used in one of the slash commands in the app.

anish-mudaraddi (approved):

PR #259, opened 2025-01-08
Promote chatops to staging

khalford:
### Description:

Taking down the ChatOps app on dev. Changing replicas to 0 and changing image tag to be placeholder for next dev changes.
Also, changing dev secrets to use the Dev Slack App rather than production which is being promoted.

**Promoting** chatops app on dev chatops cluster to staging worker cluster.

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/charts/dev/chatops/values.yaml
+++ b/charts/dev/chatops/values.yaml
@@ -1,28 +1,6 @@
 env: dev
 app: cloud-chatops
-replicaCount: 3
+replicaCount: 1
 image:
   repository: harbor.stfc.ac.uk/stfc-cloud/cloud-chatops
   tag: 5.0.2
-repos:
-  stfc:
-    - cloud-deployed-apps
-    - st2-cloud-pack
-    - cloud-grafana-dashboards
-    - cloud-pe-jupyterhub
-    - cloud-ops-tools
-    - cloud-docker-images
-    - cloud-capi-values
-    - cloud-image-builders
-    - cloud-docs
-    - cloud-openstack-horizon
-    - ansible-harbor
-    - terraform-openstack
-    - cloud-rundeck-jobs
-    - openstack-guide
-    - ansible-jupyter
-    - SCD-Openstack-Utils
-    - check-version-action
-    - cloud-helm-charts
-    - openstack-query-library
-channel: C07QX38AR5W
--- a/charts/dev/chatops/values.yaml
+++ b/charts/dev/chatops/values.yaml
@@ -1,9 +1,9 @@
 env: dev
 app: cloud-chatops
-replicaCount: 3
+replicaCount: 0
 image:
-  repository: harbor.stfc.ac.uk/stfc-cloud/cloud-chatops
-  tag: 5.0.1
+  repository: harbor.stfc.ac.uk/stfc-cloud-staging/cloud-chatops
+  tag: some-sha
 repos:
   stfc:
     - cloud-deployed-apps
@@ -25,4 +25,4 @@ repos:
     - check-version-action
     - cloud-helm-charts
     - openstack-query-library
-channel: C03RT2F6WHZ
+channel: C07QX38AR5W
--- a/charts/dev/chatops/values.yaml
+++ b/charts/dev/chatops/values.yaml
@@ -2,8 +2,8 @@ env: dev
 app: cloud-chatops
 replicaCount: 3
 image:
-  repository: harbor.stfc.ac.uk/stfc-cloud-staging/cloud-chatops
-  tag: some-sha
+  repository: harbor.stfc.ac.uk/stfc-cloud/cloud-chatops
+  tag: 5.0.2
 repos:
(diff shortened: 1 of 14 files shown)


charts/dev/chatops/values.yaml
anish-mudaraddi, line 1: just delete the chatops app entry from apps.yaml if you're done with it. 

Or better yet just delete chatops cluster folder if you're done with it

secrets/dev/cloud-chatops/apps/chatops.yaml
anish-mudaraddi, line 1: why are you reverting this back?

anish-mudaraddi (changes requested):

charts/dev/chatops/values.yaml
khalford: I would rather not delete the cluster as if I want to test out new changes I would have to create the cluster every time anyone works on it. I will comment out the app

secrets/dev/cloud-chatops/apps/chatops.yaml
khalford: Previously I was running the production ChatOps on dev as it didn't exist on Argo yet. Now the production ChatOps is moving to staging and eventually prod. These new credentials are for the **actual** dev ChatOps application. Also, I do not need every cloud members details when working in dev. Only one / my own

charts/dev/chatops/values.yaml
anish-mudaraddi: why are you updating the chart version - you're not changing anything

charts/dev/chatops/values.yaml
```diff
@@ -25,4 +25,4 @@ repos:
     - check-version-action
     - cloud-helm-charts
     - openstack-query-library
-channel: C03RT2F6WHZ
```
anish-mudaraddi, line 28: why change this?

charts/dev/chatops/values.yaml
```diff
@@ -2,8 +2,8 @@ env: dev
 app: cloud-chatops
 replicaCount: 3
 image:
-  repository: harbor.stfc.ac.uk/stfc-cloud/cloud-chatops
-  tag: 5.0.1
```
anish-mudaraddi, line 6: leave this as is - you've commented out the app, so you don't need to change this

charts/staging/chatops/values.yaml
anish-mudaraddi, line 1: staging values should match exactly what's in dev

anish-mudaraddi (changes requested):

charts/staging/chatops/values.yaml
anish-mudaraddi: the CI should pick this up but I think its broken

charts/dev/chatops/values.yaml
khalford: This is the channel for our Cloud workspace. Since going forward this will be the dev instance I have changed it to the dev workspace channel

charts/staging/chatops/values.yaml
khalford: If the values should match then why do we need different charts for dev, staging and prod?

charts/dev/chatops/values.yaml
khalford: Yeah I see what you mean

charts/dev/chatops/values.yaml
anish-mudaraddi: this is a cluster-specific value - so you need to create a new file - under clusters/dev/chatops to host this variable. 
We want to keep the shared values generic so they can easily be copied and pasted from dev -> staging without breaking anything

charts/staging/chatops/values.yaml
anish-mudaraddi: so we can test chart changes on dev, promote the generic chart to staging, and then again to prod.
to do this we must keep them as similar as possible. 

Anything that's environment/cluster specific should go in their own files that's hosted alongside the cluster

clusters/staging/worker/apps.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   project: default
   source:
     repoURL: https://github.com/stfc/cloud-deployed-apps.git
-    targetRevision: main
+    targetRevision: staging-worker-chatops
```
anish-mudaraddi, line 13: we purposefully don't use feature branches for staging - it should always follow main. The work to test and develop the chart should be done using the dev cluster - staging is for long-term testing - collecting stats for things like load and for user feedback   

Change this back to main

anish-mudaraddi (changes requested):

clusters/staging/worker/apps.yaml
khalford: Makes sense. I will do that

anish-mudaraddi (approved):
needs another reviewer since we're putting this in staging

gmatthews20 (approved):

PR #261, opened 2025-01-16
Bug: Fix longhorn settings

gmatthews20:
Longhorn settings were not being applied due to invalid values for some settings longhorn#6815
Update values to match allowed values found here: https://github.com/longhorn/charts/blob/v1.8.x/charts/longhorn/questions.yaml
Set drain policy to "block-for-eviction" to automatically evict replicas from draining nodes: https://longhorn.io/docs/1.7.2/maintenance/maintenance/#node-drain-policy-recommendations

### Special Notes:

---

### Submitter:

Have you:

---

### Reviewer

Have you:


anish-mudaraddi (approved):
thank you @gmatthews20!

PR #262, opened 2025-01-16
Promote dev capi changes to staging

anish-mudaraddi:
promote all dev changes to capi-infra chart to staging
- image changes
- bump capi-helm-chart to 0.12.2
- a few dns changes to match what's expected
- a few prometheus/alertmanager tweaks

### DIFF
--- a/clusters/staging/worker/infra-values.yaml
+++ b/clusters/staging/worker/infra-values.yaml
@@ -1,15 +1,15 @@
 openstack-cluster:
 
   controlPlane:
-    machineCount: 3
+    machineFlavor: dep-l2.tiny
 
   nodeGroups:
     - name: default-md-0
       machineCount: 3
-      machineFlavor: l3.micro
+      machineFlavor: dep-l2.tiny
 
   nodeGroupDefaults:
-    machineFlavor: l3.nano
+    machineFlavor: dep-l2.tiny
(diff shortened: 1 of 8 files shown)


clusters/staging/worker/infra-values.yaml
```diff
@@ -1,8 +1,12 @@
 openstack-cluster:
 
+  controlPlane:
+    machineCount: 3
+
   nodeGroups:
     - name: default-md-0
-      machineCount: 5
+      machineCount: 3
+      machineFlavor: l3.micro
```
DavidFair, line 10: Could we use the dep-l2 flavours as l3 is quite tight?

clusters/staging/worker/infra-values.yaml
```diff
@@ -1,8 +1,12 @@
 openstack-cluster:
 
+  controlPlane:
+    machineCount: 3
```
DavidFair, line 4: Could this be left implicit at 5, since this means we cannot lose 2 members without a full rebuild, so trying to recover becomes scarier than losing 1 and having another to start recovery from

DavidFair (changes requested):

DavidFair (approved):

gmatthews20 (approved):

PR #272, opened 2025-01-29
promote capi changes to prod

anish-mudaraddi:
bump capi to 0.12.2
bump capi addons to 0.7.1
fix alertmanager config
fix capi rollout config to not remove nodes before they are fully drained

PR #273, opened 2025-01-29
bump dev charts to staging

anish-mudaraddi:
- add opensearch chart to staging
- bump cert-manager to 1.15.3
- fix inconsistency with chatops values - `env` needs to be a cluster-specific value

This PR does not affect infrastructure charts like capi/capi-addons. See PR #272

### DIFF
--- a/charts/staging/cert-manager/Chart.yaml
+++ b/charts/staging/cert-manager/Chart.yaml
@@ -1,3 +1,8 @@
 apiVersion: v2
 name: cert-manager
 version: 1.0.0
+dependencies:
+# https://github.com/cert-manager/cert-manager/releases
+  - name: cert-manager
+    version: 1.15.3
+    repository: https://charts.jetstack.io
--- a/charts/staging/cert-manager/Chart.yaml
+++ b/charts/staging/cert-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: cert-manager
-version: 1.0.0
+version: 1.1.0
 dependencies:
 # https://github.com/cert-manager/cert-manager/releases
   - name: cert-manager
--- a/charts/staging/chatops/values.yaml
+++ b/charts/staging/chatops/values.yaml
@@ -1,4 +1,4 @@
-env: staging
+env: dev
 app: cloud-chatops
 replicaCount: 1
 image:
--- a/charts/staging/opensearch/secret-templates/opensearch.yaml
+++ b/charts/staging/opensearch/secret-templates/opensearch.yaml
@@ -1,0 +1,18 @@
+# THIS IS A TEMPLATE FOR SOPS SECRETS 
+# copy this file into relevant cluster folder in secrets directory
+# then run sops ecrypt <filename>.yaml - DON'T CHANGE THE FILENAME  
+# application: opensearch
+
+admin:
+  # Provide username and password for the initial admin user - used to configure and manage the cluster
+  # You must provide the password in plaintext and as a bycrpt hash 
+  username: admin
+  password: admin
+  # bcrypt hash of password - default is 12 rounds unless changed in settings
+  # plugins.security.password.hashing.bcrypt.rounds
+  hash: $2y$12$nfO6TYibMCV6E2wPuYoiq.psAV2dl2wmAL5ZEU.VOkaB4z4GCLVVK
+
+# IRIS IAM Client credentials
+openid:
+  clientID: iris-iam-id
+  clientSecret: iris-iam-secret
--- a/charts/staging/opensearch/secret-templates/opensearch.yaml
+++ b/charts/staging/opensearch/secret-templates/opensearch.yaml
@@ -10,7 +10,7 @@ admin:
   password: admin
   # bcrypt hash of password - default is 12 rounds unless changed in settings
   # plugins.security.password.hashing.bcrypt.rounds
-  hash: $2y$12$nfO6TYibMCV6E2wPuYoiq.psAV2dl2wmAL5ZEU.VOkaB4z4GCLVVK
+  hash: $2y$12$nfO6TYibMCV6E2wPuYoiq.psAV2dl2wmAL5ZEU.VOkaB4z4GCLVVK # admin
 
 # IRIS IAM Client credentials
 openid:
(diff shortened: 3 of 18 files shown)


charts/staging/chatops/values.yaml
```diff
@@ -1,4 +1,4 @@
-env: staging
+env: dev
```
DaveW-STFC, line 1: Think this needs to be set back to staging

DavidFair: Can we change to this something like `NotSet` and add something to the helm chart to error with something like

```
env is 'NotSet', please specify this using the override param
```

Since this is something I can see other people getting confused by

charts/staging/cert-manager/Chart.yaml
```diff
@@ -1,3 +1,8 @@
 apiVersion: v2
 name: cert-manager
 version: 1.0.0
```
DavidFair, line 3: We bumped our chart version for the other app but not this?
IMO it should be a major bump since we've changed `crds.enabled `to `installcrds` so overriding values will need to also update as appropriate

charts/staging/opensearch/secret-templates/opensearch.yaml
```diff
@@ -0,0 +1,18 @@
...
+# application: opensearch
+
+admin:
+  # Provide username and password for the initial admin user - used to configure and manage the cluster
+  # You must provide the password in plaintext and as a bycrpt hash 
+  username: admin
+  password: admin
+  # bcrypt hash of password - default is 12 rounds unless changed in settings
+  # plugins.security.password.hashing.bcrypt.rounds
+  hash: $2y$12$nfO6TYibMCV6E2wPuYoiq.psAV2dl2wmAL5ZEU.VOkaB4z4GCLVVK
```
DavidFair, line 13: Is this `admin` as a hash too? We should add a comment if it is stating that so it's obvious

charts/staging/opensearch/secret-templates/opensearch.yaml
```diff
@@ -0,0 +1,18 @@
+# THIS IS A TEMPLATE FOR SOPS SECRETS 
+# copy this file into relevant cluster folder in secrets directory
+# then run sops ecrypt <filename>.yaml - DON'T CHANGE THE FILENAME
```
DavidFair, line 3: I'm not super keen on this flow, as forgetting to run `sops encrypt` (also typo) is a foot gun

Could we say:
```
# copy this file out of the git repo, run sops encrypt
# then copy the encrypted file ONLY back to the relevant cluster's secrets directory
```

charts/staging/opensearch/templates/admin-secret.yaml
DavidFair, line 1: This should be the template we give users and SOPS encrypt rather than using a values file

A values file will appear in the helm history, or `get values` leaking secrets, or even ArgoCD UI. However, a secret that's decrypted and applied directly (if possible using helm-secrets) wouldn't appear unless someone ran `helm template` (which means they have the secret) or `kubectl describe secret` (which can be subject to an RBAC)

charts/staging/opensearch/templates/longhorn-storageclass.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: longhorn-opensearch
+provisioner: driver.longhorn.io
+parameters:
+  numberOfReplicas: "1"
```
DavidFair, line 7: This should be `>1` by default, and safe by default or opt out if people don't expect resiliency 
Otherwise losing a single node == data loss in our default config

DavidFair (changes requested):

charts/staging/chatops/values.yaml
anish-mudaraddi: leave it for another PR - @khalford should look at this

charts/staging/opensearch/secret-templates/opensearch.yaml
anish-mudaraddi: I'm concerned about PR scope - this is just moving working config into staging
Also I'm keen on moving all of the charts over to cloud-helm-charts - I'll revisit this later

charts/staging/opensearch/templates/longhorn-storageclass.yaml
anish-mudaraddi: no - opensearch already copies indexes across nodes - adding more than 1 longhorn replicas means duplicating it twice

charts/staging/opensearch/templates/admin-secret.yaml
anish-mudaraddi: I deliberately added this here because then we _could_ specify secrets via helm values. It removes the post-deployment step of having to manually create this secret. Even if we were to encrypt this secret and run a simple `kubectl apply -f` we lose the ability to manage this secret via argocd. 

I'd be more than happy to remove this if we ever get ArgoCD Vault Plugin or K8s Secrets Operator running which can grab secrets from Hashicorp Vault/Keeper respectively

charts/staging/cert-manager/Chart.yaml
anish-mudaraddi: why do we care about versioning these charts? 
They are only used by argocd

charts/staging/opensearch/secret-templates/opensearch.yaml
anish-mudaraddi: yes its admin as a bcrypt hash

charts/staging/opensearch/secret-templates/opensearch.yaml
anish-mudaraddi: but how can we encrypt the file without being in the correct repo with the relevant `.sops.yaml`

charts/staging/opensearch/templates/admin-secret.yaml
DavidFair: Ah okay I thought helm secrets would allow us to do this too
Please feel free to disregard

DavidFair (approved):

DaveW-STFC (approved):

PR #274, opened 2025-01-29
bump to latest capi image on dev

anish-mudaraddi:
dev now has the latest capi images

### DIFF
--- a/charts/dev/capi-infra/values.yaml
+++ b/charts/dev/capi-infra/values.yaml
@@ -1,6 +1,6 @@
 openstack-cluster:
   kubernetesVersion: "1.31.4"
-  machineImage: "capi-ubuntu-2204-kube-v1.31.4-2025-01-07"
+  machineImage: "capi-ubuntu-2204-kube-v1.31.4-2025-01-24"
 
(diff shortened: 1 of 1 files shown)


charts/dev/capi-infra/values.yaml
```diff
@@ -354,19 +354,6 @@ openstack-cluster:
...
-        # The maximum number of node group machines that can be unavailable during the update
-        # Can be an absolute number or a percentage of the desired count
-        maxUnavailable: 0
-        # The maximum number of machines that can be scheduled above the desired count for
-        # the group during an update
-        # Can be an absolute number or a percentage of the desired count
-        maxSurge: 1
-        # One of Random, Newest, Oldest
-        deletePolicy: Random
-
```
gmatthews20, line 369: Why is the rolloutStrategy being removed?

gmatthews20 (changes requested):

charts/dev/capi-infra/values.yaml
anish-mudaraddi: bug - fixed

gmatthews20 (approved):

PR #359, opened 2025-07-22
Move Opensearch deployment onto Staging + add permissions for opensearch

anish-mudaraddi:
we've tested these since we've made these available via cloud-helm-charts

### DIFF
--- a/charts/dev/argocd/Chart.yaml
+++ b/charts/dev/argocd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: argocd
-version: 1.0.2
+version: 1.0.3
 dependencies:
   # https://github.com/argoproj/argo-helm/releases
   - name: argo-cd
--- a/charts/prod/argocd/Chart.yaml
+++ b/charts/prod/argocd/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: argocd
-version: 1.0.1
+version: 1.0.2
 dependencies:
   # https://github.com/argoproj/argo-helm/releases
   - name: argo-cd
-    version: 7.8.0
+    version: 8.0.14
     repository: https://argoproj.github.io/argo-helm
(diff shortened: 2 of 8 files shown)


.github/workflows/pr_copied_files.yaml
```diff
@@ -18,19 +18,6 @@ jobs:
...
-          for file in ${{ steps.changed-files.outputs.staging_all_changed_files }}; do
-            echo "Checking $file..."
-            if ! ${{ github.workspace }}/.github/workflows/pr_check_copies.sh "$file" "staging" "dev"; then
-              echo "❌ $file differs from dev version"
-              exit 1
-            fi
-          done
-          echo "✅ All staging files match dev"
-
       - name: Check prod files
```
DavidFair, line 21: Nit: Can we rename this pipeline / job to something like `Check prod is promotion from staging` now that we're only checking promotions from staging

charts/dev/capi-infra/Chart.yaml
DavidFair, line 1: Can we drop all chart updates from this PR? Since the CI will automatically open PR(s) for these anyway

That way we can land the changes for moving OpenSearch and adding perms independently, as the chart versions have since updated

clusters/staging/worker/opensearch-values.yaml
```diff
@@ -0,0 +1,344 @@
...
+                config:
+                  subject_key: preferred_username
+                  roles_key: groups
+                  openid_connect_url: "https://iris-iam.stfc.ac.uk/.well-known/openid-configuration"
+                  enable_ssl: true
+                  verify_hostnames: true
+              authentication_backend:
+                type: noop
+           
+    roles.yml: |
```
DavidFair, line 115: I think these are the same in all versions looking at it, we should move it into the chart rather than have it duplicated in all environments?

DavidFair (changes requested):

PR #363, opened 2025-07-29
Promote argocd again

DavidFair:
### Description:

Take 2 on promoting our version from dev (which has been running fine for a while) to staging and prod to get it to a known good version
Also bump dev (only) to latest argocd to let that start baking, so we can promote that later on too

---

### Submitter:

Have you:

---

### Reviewer

Have you:


DavidFair:
Failure expected, since we're promoting to two environments at once
We'll need a 2 person review for this one

AlexCK-STFC (approved):

PR #365, opened 2025-07-31
scale up rabbit-consumers

anish-mudaraddi:
scale up rabbit consumer deployment from 1 to 3


jacob-ward (approved):
Can confirm 3 > 1 - approved

AlexCK-STFC (approved):

PR #390, opened 2025-09-05
bump opensearch storage and lower replicas

anish-mudaraddi:
bump the cinder storage for opensearch - we're already reaching the cap after 35 days


AlexCK-STFC:
@anish-mudaraddi Do we wanna set up log rotation? It's not that bad to set up, I ran tests previously
Or is storage cheap enough we don't care

anish-mudaraddi:
yes we should setup log-rotation - for another PR

PR #532, opened 2026-04-13
Match our staging and prod rabbit values

DavidFair:
### Description:

~The container is not starting, I suspect this is because the replica count isn't set properly, but to start any further troubleshooting match these values to a known working copy~

Adjust the layout to match our prod and staging values to make a diff easier too

---

### Submitter:

Have you:

---

### Reviewer

Have you:

### DIFF
--- a/clusters/staging/worker/rabbit-consumer-values.yaml
+++ b/clusters/staging/worker/rabbit-consumer-values.yaml
@@ -1,14 +1,12 @@
-
 stfc-cloud-rabbit-consumer:
   consumer:
     logLevel: INFO
+    aquilon:
+      defaultPrefix: vm-openstack-Dev-
 
     queues:
       - "ral.info"
       - "ral.error"
-
-    aquilon:
-      defaultPrefix: vm-openstack-Dev-
 
     rabbitmq:
       host: dev-service4.nubes.rl.ac.uk
--- a/clusters/staging/worker/rabbit-consumer-values.yaml
+++ b/clusters/staging/worker/rabbit-consumer-values.yaml
@@ -17,4 +17,4 @@ stfc-cloud-rabbit-consumer:
       projectId: c9aee696c4b54f12a645af2c951327dc
 
   kerberosSidecar:
-    principle: "HTTP/service1.nubes.rl.ac.uk"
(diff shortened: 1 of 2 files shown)


clusters/staging/worker/rabbit-consumer-values.yaml
```diff
@@ -1,15 +1,14 @@
-
 stfc-cloud-rabbit-consumer:
+  replicaCount: 1
```
anish-mudaraddi: its already set to 1 - https://github.com/stfc/cloud-helm-charts/blob/main/charts/stfc-cloud-rabbit-consumer/values.yaml

you might need to change this file to reflect: https://github.com/stfc/cloud-helm-charts/blob/main/charts/stfc-cloud-rabbit-consumer/dev-values.yaml - which we seem to have not done


```
stfc-cloud-rabbit-consumer:
  consumer:
    logLevel: INFO
    image:
      repository: rabbit-consumer
      tag: local
      # This allows us to build the image locally and use it in the chart
      pullPolicy: Never
  
    aquilon:
      defaultPrefix: vm-openstack-Dev-
  
    rabbitmq:
      host: dev-openstack.stfc.ac.uk
  
    openstack:
      authUrl: https://dev-openstack.stfc.ac.uk:5000/v3
      computeUrl: https://dev-openstack.stfc.ac.uk:8774/v2.1
      projectId: c9aee696c4b54f12a645af2c951327dc
  
  kerberosSidecar:
    principle: "HTTP/dev-service1.nubes.rl.ac.uk"

```

anish-mudaraddi (changes requested):

clusters/staging/worker/rabbit-consumer-values.yaml
DavidFair: Looking at the chart in staging/rabbit-consumer/chart.yml it's still pulling in as `stfc-cloud-rabbit-consumer` unless we need to match the name in the child of this chart?

I'll drop the 1 from both staging and prod to unify them on that then

anish-mudaraddi (dismissed):

meoflynn (approved):

anish-mudaraddi (approved):

